PR #1, opened 2023-06-12
Initial cmake configuation and C skeletons

bgeihe:


cmake/FindJulia.cmake
```diff
@@ -0,0 +1,119 @@
...
+#
+# Once julia is found, use julia-config
+#
+execute_process(
+    COMMAND ${JULIA_EXECUTABLE} -e "print(dirname(Sys.BINDIR))"
+    OUTPUT_VARIABLE JULIA_HOME
+    RESULT_VARIABLE RESULT
+)
+if( NOT RESULT EQUAL 0 )
+    message( WARNING "Could not determine julia's home directory" )
```
sloede, line 30:
```suggestion
    message( WARNING "Could not determine julia's bin directory" )
```
?

cmake/FindJulia.cmake
```diff
@@ -0,0 +1,119 @@
...
+#
+execute_process(
+    COMMAND ${JULIA_SHARE}/julia-config.jl --cflags
+    OUTPUT_VARIABLE JULIA_CFLAGS
+    RESULT_VARIABLE RESULT
+)
+
+execute_process(
+    COMMAND ${JULIA_SHARE}/julia-config.jl --ldflags
+    OUTPUT_VARIABLE JULIA_LDLAGS
```
sloede, line 55:
```suggestion
    OUTPUT_VARIABLE JULIA_LDFLAGS
```

src/libtrixi.h
```diff
@@ -0,0 +1,11 @@
+#ifndef LIBTRIXI_H_
+#define LIBTRIXI_H_
+
+#include <mpi.h> // required for MPI
+
+void trixi_initialize();
```
sloede, line 6:
```suggestion
void trixi_initialize(MPI_Fint* comm);
```
To ,act function definition?

src/libtrixi.h
sloede, line 1: Maybe 🚲🛖, but AFAICT, most libraries just call their main file as `name.c/h`, not `libname.c/h`, e.g., [`netcdf.h`](https://github.com/Unidata/netcdf-c/blob/main/include/netcdf.h) or [`curl/curl.h`](https://github.com/curl/curl/blob/master/include/curl/curl.h). Thus I'd just call it `trixi.h` and `trixi.c`, but of course this can also be decided later.

sloede (commented):
Great that you already got started! I've had a quick first look and commented on a few things...

bgeihe:
Do you think this could be merged @sloede, such that the MESSy people can start using it?

cmake/FindJulia.cmake
sloede, line 1: Is this file based on an example or did you write it by yourself from scratch? If it is the former, it would be great to include a link to the source.

CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+# Specify the minimum version.
+cmake_minimum_required ( VERSION 3.0 )
+
+# Specify the project info.
+project ( trixi VERSION 0.1.0 DESCRIPTION "Interface library for using Trixi.jl from C/C++/Fortran" )
+
+# Enable fortran
+enable_language(Fortran)
```
sloede, line 8: Do we also need to enable C? Or, for clarity, should we just include it even if it is enabled automatically to make it explicit?

src/trixi.c
```diff
@@ -0,0 +1,83 @@
...
+    return 1.0;
+}
+
+
+void trixi_integrate() {
+
+}
+
+
+void julia_eval_repl(const char * code) {
```
sloede, line 54:
```suggestion
void julia_eval_string(const char * code) {
```
I wouldn’t call it REPL since technically it is not a REPL, and it behaves a little different.

src/trixi.c
```diff
@@ -0,0 +1,83 @@
...
+
+
+double trixi_get_timestep() {
+
+    return 1.0;
+}
+
+
+void trixi_integrate() {
+
```
sloede, line 50:
```suggestion
    printf(…)
```
Maybe add a dummy printout just for testing purposes?

src/trixi.h
```diff
@@ -0,0 +1,12 @@
+#ifndef LIBTRIXI_H_
```
sloede, line 1:
```suggestion
#ifndef TRIXI_H_
```
To match filename (needs to be changed below as well)

CMakeLists.txt
sloede, line 1: Right now, both the library and the example program depend on an appropriate MPI installation being present at the default location in the system. For portability, I think you also need to use FindMPI and pass the include dirs and libraries to the respective targets. TBH, I’m surprised it even compiles as it is 😅

src/examples/fortran_hello_world.f90
```diff
@@ -0,0 +1,46 @@
+module LibTrixi
```
sloede, line 1: I think the module that describes the C interface should go into its own file, such that it is usable outside the example.

CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+
+
+
+# Find MPI
+find_package( MPI REQUIRED )
+
+
+
+# Build examples
+add_executable(fortran_hello_world src/examples/fortran_hello_world.f90)
```
sloede, line 48: I recommend to not put the examples folder inside the src dir but as a new toplevel directory in the repo. Then, the examples folder should get its on CMakeLists.txt and this stuff here should be removed. That way, you have complete isolation of the library from the examples. This would be a great help for others who want to use the Trixi.jl library, since they now can see what they need to do to link their own program to our library. That is,they can use your example as a blueprint.

sloede (changes requested):
I think this is a great first example Setup! My main concern currently is that it doesn’t explicitly have MPI as a dependency in CMake. You could either fix CMake or remove MPI entirely (for now; it isn’t really tested yet anyways). Also, the separation of library and examples.

Also, I strongly recommend to add some short sections to the README.md, eg, an installation and usage section.

cmake/FindJulia.cmake
bgeihe: Indeed, I started off with some snippets but in the end rewrote everything to make use of `julia-config.jl`

CMakeLists.txt
bgeihe: I will add it.

CMakeLists.txt
bgeihe: This is actually done right now. What confused me was that e.g. `gfortran` was used and not `mpifr` . But this seems to be the way to go. Anyways it works with `mpirun`.

sloede: Ah yes, I got confused b/c I looked at it at my phone. `MPI` actually loaded as a requirement and thus it makes sense that it compiles ok 👍 However, as you pointed out, usually you need `mpif90` for all headers, libs etc. to work - however, since we have it all in a default location on roci, maybe it just works out of the box 🤷 anyhow, closing this now convo

sloede:
FYI, right now on roci it does not compile out of the box: When I run `cmake . && make` from the root dir, I get
```
Scanning dependencies of target trixi
[ 25%] Building C object CMakeFiles/trixi.dir/src/trixi.c.o
/home/mschlott/hackathon/libtrixi/src/trixi.c:3:10: fatal error: mpi.h: No such file or directory
    3 | #include <mpi.h>
      |          ^~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/trixi.dir/build.make:63: CMakeFiles/trixi.dir/src/trixi.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:105: CMakeFiles/trixi.dir/all] Error 2
make: *** [Makefile:84: all] Error 2
```
If I try to be smart about it and use `CC=mpicc F90=mpif90 cmake . && make` instead, I get a little further but not super far:
```
Scanning dependencies of target trixi
[ 25%] Building C object CMakeFiles/trixi.dir/src/trixi.c.o
[ 50%] Linking C shared library libtrixi.so
[ 50%] Built target trixi
Scanning dependencies of target fortran_hello_world
[ 75%] Building Fortran object CMakeFiles/fortran_hello_world.dir/src/examples/fortran_hello_world.f90.o
[100%] Linking Fortran executable fortran_hello_world
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_stderr_stream'
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_call2'
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_symbol'
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_base_module'
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_eval_string'
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_exception_occurred'
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_atexit_hook'
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_printf'
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_get_global'
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_init'
/usr/bin/ld: libtrixi.so.0.1.0: undefined reference to `jl_stderr_obj'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/fortran_hello_world.dir/build.make:90: fortran_hello_world] Error 1
make[1]: *** [CMakeFiles/Makefile2:78: CMakeFiles/fortran_hello_world.dir/all] Error 2
make: *** [Makefile:84: all] Error 2
```

sloede:
I fixed the `undefined reference` errors from above in aeb2612 by reordering the linked libraries such that the symbols in `libjulia.so` are picked up properly.

bgeihe:
Got your point!
So, should we just add an example `CMakeLists.txt` which compiles the `fortran_hello_world` as a standalone cmake project. Or maybe just add the cmake code to `README.md`?

PR #2, opened 2023-06-13
Add initial version of a LibTrixi.jl package

sloede:
The idea is to collect all relevant interface functionality on the Julia side here in a local package instead of having to put it in Trixi.jl directly. Everything in here very much WIP!


sloede:
I improved upon `trixi_initialize` by allowing one to pass a so-called `libelixir` that will set up the corresponding simulation similarly to as you would do it in a regular Trixi.jl elixir, but not actually run the simulation yet.

Thus, I think this is ready for a review.

@bgeihe I think it would be great if you can have a look and try to see if you spot anything weird/problematic, or just let me know if you have any questions. Otherwise, since this is not yet really used in anything resembling production, you can also just merge it (please select "Squash and merge" since I had a lot of stupid commits).

LibTrixi.jl/src/api_c.jl
```diff
@@ -0,0 +1,131 @@
...
+    delete_simstate!(simstate_handle)
+    GC.gc()
+
+    return nothing
+end
+
+trixi_finalize_cfptr() = @cfunction(trixi_finalize, Cvoid, (Ptr{Cvoid},))
+
+"""
+    trixi_calculate_dt(simstate_handle::Cint))::Cdouble
```
bgeihe, line 86:
```suggestion
    trixi_calculate_dt(simstate_handle::Cint)::Cdouble
```

LibTrixi.jl/src/simulationstate.jl
```diff
@@ -0,0 +1,58 @@
...
+    integrator::IntegratorType
+end
+
+# Global variables to store different simulation states
+# This allows one to handle multiple Trixi.jl simulations independently
+#
+# Opaque handle type that can be passed to and stored in the C program
+const SimulationStateHandle = Cint
+# Variable that internally holds different simulation states such that they are garbage
+# collected prematurely
```
bgeihe, line 14:
```suggestion
# Variable that internally holds different simulation states such that they are not garbage
# collected prematurely
```
?

LibTrixi.jl/src/simulationstate.jl
```diff
@@ -0,0 +1,58 @@
...
+function store_simstate(simstate)
+    if simstate_counter[] >= typemax(SimulationStateHandle)
+        error("maximum number of storable simulation states reached: ",
+              typemax(SimulationStateHandle))
+    end
+
+    simstate_counter[] += 1
+    handle = simstate_counter[]
+    if in(handle, keys(simstates))
+        error("simstate handle already exists - this should be possible...")
```
bgeihe, line 30:
```suggestion
        error("simstate handle already exists - this should not be possible...")
```

bgeihe (commented):
Nice! Works for me as well.

sloede:
@bgeihe Thanks for the review! I applied all your suggestions. Is this good to merge from your POV?

Issue #6, opened 2023-06-21
Next steps


sloede:
- [x] #7
- [x] #3
- [x] #8
- [x] #9
- [x] #17
- [x] #15
- [x] #10
- [x] #31
- [x] #11

For the future (neither time nor mission critical at the moment):
- [x] #22

sloede:
All TODOs have been checked 🥳 We can revive this issue of it becomes necessary again.

PR #14, opened 2023-06-21
Initial coupling of libtrixi (C) and LibTrixi.jl

### LINKED ISSUE #7: Couple libtrixi to LibTrixi.jl
The goal is to be able to control the currently implemented dummy simulation in LibTrixi.jl from our example Fortran program.

### LINKED ISSUE #8: Separate init functions for general startup and simulation setup
Right now, `trixi_initialize` in LibTrixi.jl sets up a concrete simulation. However, it might be useful/necessary to separate this into two functions:
* One function is only ever called once and does all "global" initialization of libtrixi/Trixi.jl (e.g., this could continue to be named `trixi_initialize`)
* One function is called to create a concrete new simulation setup (e.g., this could be named `trixi_create_simulation` or something similar)

bgeihe:
This resolves #7 and resolves #8


LibTrixi.jl/src/api_jl.jl
```diff
@@ -1,4 +1,4 @@
-function trixi_initialize_jl(filename)
+function trixi_setup_simulation_jl(filename)
```
sloede, line 1: It's a little 🚲🛖 (bike shedding), but "setup" imho is a noun, not a verb. Thus it should be either something like
```suggestion
function trixi_create_simulation_jl(filename)
```
or at least
```suggestion
function trixi_set_up_simulation_jl(filename)
```
If you ask me, the latter one has too many underscores, thus I'd personally prefer the first version (or a variant thereof).

Note that if this is changed here, it should be changed consistently for all similarly-named functions

examples/CMakeLists.txt
```diff
@@ -13,3 +13,19 @@ target_include_directories(fortran_hello_world PRIVATE ${MPI_Fortran_INCLUDE_DIR
...
 install( TARGETS fortran_hello_world )
+
+
+
+#
+# Simple Trixi Controller
+#
+add_executable(simple_trixi_controller simple_trixi_controller.c)
+
+target_link_libraries(simple_trixi_controller PRIVATE ${PROJECT_NAME})
```
sloede, line 24: I don't understand - how is `${PROJECT_NAME}` set? Or this this CMake file still just included by the main CMake file?

sloede: Ah, yes, I just checked, it is :-) NVM!

src/trixi.f90
sloede, line 1: This Fortran binding file is very outdated by now... maybe create an issue to update it and add it to #6?

src/trixi.f90
```diff
@@ -10,7 +10,7 @@ subroutine trixi_initialize(comm) bind(c)
     subroutine trixi_finalize() bind(c)
     end subroutine
 
-    subroutine trixi_integrate() bind(c)
+    subroutine trixi_step() bind(c)
     end subroutine
```
sloede, line 14: Shouldn't this be something like
```suggestion
    subroutine trixi_step(handle) bind(c)
      use, intrinsic :: iso_c_binding, only: c_int
      integer, intent(in) :: handle
    end subroutine
```
?

examples/simple_trixi_controller.c
```diff
@@ -0,0 +1,31 @@
...
+
+    // Setup the Trixi simulation
+    // We get a handle to use subsequently
+    int handle = trixi_setup_simulation( elixir );
+
+    // Get time step length
+    printf("Current time step length: %f\n", trixi_calculate_dt(handle));
+
+    // Main loop
+    while ( trixi_is_finished( handle ) == 0 ) {
```
sloede, line 21:
```suggestion
    while ( !trixi_is_finished( handle ) ) {
```
Since we are in C world here, we can just use the canonical "0 is equal to false". At least IMHO this is more descriptive

examples/simple_trixi_controller.c
```diff
@@ -0,0 +1,31 @@
...
+
+        trixi_step( handle );
+    }
+
+    // Finalize Trixi
+    trixi_finalize( handle );
```
sloede, line 27: I just realize: If we have something like `trixi_initialize_simulation()` which returns a `handle`, we should also have a matching `trixi_finalize_simulation(handle)`. Then, for complete "tear down", we can again use `trixi_finalize()` without any arguments

sloede (changes requested):
This is shaping up really nicely! I've left a few comments and suggestions, but overall great work 💪

src/trixi.c
```diff
@@ -13,47 +13,105 @@ static jl_value_t* checked_eval_string(const char* code, const char* func, const
...
+ *
+ *  \return handle (integer) to Trixi simulation instance
+ */
+int trixi_setup_simulation(const char * elixir) {
+
+    // Get function pointer
+    int (*trixi_setup_simulation_c)(const char *) = jl_unbox_voidpointer( checked_eval_string("trixi_setup_simulation_cfptr()", LOC) ) ;
+
+    // Call function
+    return trixi_setup_simulation_c( elixir );
```
sloede, line 46:
```suggestion
int trixi_setup_simulation(const char * libelixir) {

    // Get function pointer
    int (*trixi_setup_simulation_c)(const char *) = jl_unbox_voidpointer( checked_eval_string("trixi_setup_simulation_cfptr()", LOC) ) ;

    // Call function
    return trixi_setup_simulation_c( libelixir );
```
I'd call it `libelixir` (or we can come up with another name) to distinguish it from a regular elixir

sloede (dismissed):
LGTM! I suggest to merge #16 in here first and then everything to `main`

PR #18, opened 2023-06-22
Store function pointers in library

### LINKED ISSUE #9: Store C function pointers from LibTrixi.jl inside libtrixi
Motivated by a separation of concerns and - possibly - by performance considerations, we should store the C function pointers inside libtrixi instead of getting them each time with a call to `jl_eval_string("some_function_cfptr()")`. Ideally, we'd only have calls to low-level libjulia functions in `trixi_initialize` and `trixi_finalize`, while all others functions directly call into LibTrixi.jl using C function pointers.

sloede:
Closes #9.

### DIFF
--- a/src/trixi.c
+++ b/src/trixi.c
@@ -5,11 +5,35 @@
 #include "trixi.h"
 
 
-
+// Auxiliary declarations for more helpful error messages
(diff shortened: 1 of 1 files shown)


bgeihe:
Nice!

PR #19, opened 2023-06-22
Add initial CI testing capabilities

### LINKED ISSUE #17: Add tests to the repo
At the very least, we should check that
* everything compiles
* the example programs run without errors

In the future, we should of course also add numeric tests that involve the results from running a Trixi.jl simulation through one of our example programs, but that's something for another time...

sloede:
Resolves #17.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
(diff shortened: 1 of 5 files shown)


sloede:
It would be great if you could have a quick look @bgeihe such that you have an idea of what I changed and did here. Right now, we are only doing smoke tests, i.e., I am running the simple example and verify that it does not return with a non-zero exit code.

sloede:
Oh, and coverage is only covering the non-Julia files at the moment, so this is something to be investigated in the future.

bgeihe:
Something great again! 

Concerning the coding, I'm lost though.

PR #20, opened 2023-06-22
Add Trixi.jl and OrdinaryDiffEq.jl

bgeihe:
Just added 
```
[deps]
OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
Trixi = "a7f1ee26-1774-49b1-8366-f1abc58fbfcb"
```
to `Project.toml`. Is there anything else to do?

Note that I disabled MPI for the time being as somehow my program exited early else.


sloede:
> Note that I disabled MPI for the time being as somehow my program exited early else.

Can you post the complete output/error message/



> Is there anything else to do?

Yes, please also add a `[compat]` entry for the current versions of Trixi.jl/OrdinaryDiffEq.jl.

LibTrixi.jl/Project.toml
```diff
@@ -3,5 +3,10 @@ uuid = "7e097bd5-a775-4bc1-90b6-ad92bd7220c1"
 authors = ["Michael Schlottke-Lakemper <michael@sloede.com>", "Benedict Geihe <bgeihe@uni-koeln.de>"]
 version = "0.1.0"
 
+[deps]
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
+Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+Trixi = "a7f1ee26-1774-49b1-8366-f1abc58fbfcb"
```
sloede, line 9: These all require a corresponding `[compat]` entry, similar to
https://github.com/trixi-framework/Trixi.jl/blob/bea4bfed27ce896edcfa3dbe79861a4faef4dec7/Project.toml#L51-L53

LibTrixi.jl/src/LibTrixi.jl
```diff
@@ -1,5 +1,10 @@
 module LibTrixi
 
+using Reexport: @reexport
+
+@reexport using OrdinaryDiffEq
+@reexport using Trixi
```
sloede, line 6: We have discussed this for TrixiShallowWater.jl before, and concluded that it is probably _not_ a good idea to reexport everything from those packages. Why? Because whenever there is a breaking change in those packages, we also need to do a breaking change in LibTrixi.jl. Thus I'd recommend to just write
```suggestion
using OrdinaryDiffEq
using Trixi
```
or similar...

LibTrixi.jl/src/api_jl.jl
```diff
@@ -12,59 +12,32 @@ function trixi_initialize_simulation_jl(filename)
 end
 
 function trixi_finalize_simulation_jl(simstate)
-    (; u0, u) = simstate.integrator
-
-    # Resize arrays to zero length
-    resize!(u0, 0)
-    resize!(u, 0)
+    # TODO: call the SummaryCallback(), but its position is arbitrary
+    simstate.integrator.opts.callback.discrete_callbacks[1]()
```
sloede, line 16: You could loop over the `discrete_callbacks` and then check if `discrete_callbacks[i].affect! isa SummaryCallback` or something similar.

LibTrixi.jl/src/LibTrixi.jl
bgeihe: Sounds reasonable!

How do I access e.g. `LinearScalarAdvectionEquation1D` in the libelixir then?

LibTrixi.jl/src/api_jl.jl
bgeihe: Will try!

bgeihe:
> > Note that I disabled MPI for the time being as somehow my program exited early else.
> 
> Can you post the complete output/error message/

Unfortunately there is none. It just exists when `using Trixi` is evaĺuated with exit code 15.

LibTrixi.jl/src/LibTrixi.jl
sloede: Good point. We could either a) have them exposed through `using LibTrixi.Trixi` or b) add them directly to the project file in `libtrixi-julia`. I think b) is the more flexible option, since that will allow us to directly control exactly which version of Trixi.jl and OrdinaryDiffEq.jl is used.

sloede:
> > > Note that I disabled MPI for the time being as somehow my program exited early else.
> > 
> > 
> > Can you post the complete output/error message/
> 
> Unfortunately there is none. It just exists when `using Trixi` is evaĺuated with exit code 15.

Hm, this is annoying. I already noticed that this is one of the downsides of using Julia in such a way, that the error messages are less helpful :-/

sloede:
The error you see in CI seems to be related to how you access the time step size:
```
ErrorException("type NamedTuple has no field dtpropose")
```
in https://github.com/trixi-framework/libtrixi/actions/runs/5348219819/jobs/9697806606?pr=20#step:11:16

sloede:
> > > > Note that I disabled MPI for the time being as somehow my program exited early else.
> > > 
> > > 
> > > Can you post the complete output/error message/
> > 
> > 
> > Unfortunately there is none. It just exists when `using Trixi` is evaĺuated with exit code 15.
> 
> Hm, this is annoying. I already noticed that this is one of the downsides of using Julia in such a way, that the error messages are less helpful :-/

Also note that you can first try to get everything up and running without using C at all - just directly load LibTrixi.jl in a Julia REPL and use the same interface. That should be much easier to figure out the whole setup, and once that works, it _should_ just work directly from C/Fortran as well (famous last words)

LibTrixi.jl/src/LibTrixi.jl
bgeihe: Still unsure about this: should we always start the libelixir files with
`using Trixi, OrdinaryDiffEq`
or can this be done somewhere else, s.t. one can still write e.g. `LinearScalarAdvectionEquation1D` in the libelixir?

bgeihe:
In the current state of the branch, the MPI problem is gone. I will accept this without asking questions.

LibTrixi.jl/src/LibTrixi.jl
```diff
@@ -1,5 +1,8 @@
 module LibTrixi
 
+using OrdinaryDiffEq
+using Trixi
```
sloede, line 4: We usually frown upon just blindly `using SomePackage` since that pulls in everything from that namespace that has been `export`ed from said package, and thus it is not possible/easy to decide anymore from which package a particular function is (at least not by just looking at the source code).
Thus, I propose to change this part of the code as follows:
```suggestion
using OrdinaryDiffEq: OrdinaryDiffEq, step!, check_error
using Trixi: Trixi
```

LibTrixi.jl/src/LibTrixi.jl
sloede: TBH, I wouldn't do it anywhere else. The advantage of having it in the libelixir itself is that it makes it more flexible (we could also add additional packages that we don't even know of yet. Also, if the packages are included, it makes it a proper Julia file in its own right, i.e., something one can run in an arbitrary Julia session for testing etc.

LibTrixi.jl/src/api_jl.jl
sloede: It was a little different than I thought, but this is now fixed with 135b362.

LibTrixi.jl/src/LibTrixi.jl
sloede: I resolved this by adding Trixi.jl and OrdinaryDiffEq.jl as explicit dependencies to the project directory in 799859c. That way, one can just write
```julia
using LibTrixi
using OrdinaryDiffEq
using Trixi
```
inside the libelixirs and be done with it. That allows one also to easily switch out the Trixi.jl package for a local installation/dev'd version if desired.

bgeihe:
:smiley: All this looks very nice! The new install script also ran flawlessly. 

What is missing for merging this? 

- [ ] Absolute paths in `LocalPreferences.toml`: When relative paths were used during setup and do not match with the run directory used later
- [ ] Remove fortran_hello_world? Is this still useful?
- [ ] Remove `libelixir_demo.jl`? This would require a mockup of Trixi and OrdinaryDiffEq. Does this still make sense, now that we have Trixi and OrdinaryDiffEq included anyway

sloede:
> 😃 All this looks very nice! The new install script also ran flawlessly.

Great to hear 💪

> Absolute paths in `LocalPreferences.toml`: When relative paths were used during setup and do not match with the run directory used later

Not sure I understand what you mean. My `LocalPreferences.toml` looks as follows:
```toml
[MPIPreferences]
_format = "1.0"
abi = "OpenMPI"
binary = "system"
libmpi = "/lib/x86_64-linux-gnu/libmpi.so"
mpiexec = "mpiexec"

[P4est]
libp4est = "/opt/p4est/2.8.5/lib/libp4est.so"
libpsc = "/opt/p4est/2.8.5/lib/libsc.so"
```
Thus, there are no relative paths anymore. Do you think that's a problem? I'd say for robustness against accidentally using the wrong library, it's actually a feature 😬

> Remove fortran_hello_world? Is this still useful?

Yes, I think we should remove it, now that setting up everything works _relatively_ smoothly.

> Remove `libelixir_demo.jl`? This would require a mockup of Trixi and OrdinaryDiffEq. Does this still make sense, now that we have Trixi and OrdinaryDiffEq included anyway?

Yes, let's remove it 👍

In general, I agree, with these points fixed we should merge this PR as soon as possible to be able to proceed with the next steps.

bgeihe:
Mine was
```
[P4est]
libp4est = "../p4est/lib/libp4est.so"
libpsc = "../p4est/lib/libsc.so"
```
as I had initially used `--p4est-library ../p4est/lib/libp4est.so `

I hope 71eda6ad7f2cc3a2a729056c5aa39f8c9379e7aa fixes this.

bgeihe:
Can we also remove `LibTirix.jl/README.md`? The example would not work anymore and the julia way is meanwhile explained in root README.md

sloede:
I updated the README in LibTrixi.jl in ed1b23c. If it's ok with you, I'd like to keep it for now.

sloede:
Removed `libelixir_demo.jl` in 07f5837.

sloede (approved):
This looks ready to merge for me!

sloede (approved):
LGTM!

sloede (approved):

PR #24, opened 2023-06-23
Adapt julia install script to account for Trixi, p4est and system MPI

bgeihe:
Found the `strict` argument to `Pkg.precompile` here: https://github.com/JuliaLang/Pkg.jl/pull/2362


sloede:
This PR seems to partially contain also the changes in #20. Is #20 obsolete or which one is the "current" working PR?

sloede:
I have updated the init script in d73f6e7 (I hope that's ok). It is now a little more flexible and (hopefully) robust, and tries to take off as much work from the user as possible. Note the new usage (with `-h`):
```
usage: libtrixi-init-julia [-h] [-f]
                           [--julia-depot JULIA_DEPOT]
                           [--julia-exec JULIA_EXEC]
                           [--mpi-library MPI_LIBRARY]
                           [--p4est-library P4EST_LIBRARY]
                           [--sc-library P4EST_LIBRARY]
                           PREFIX_OR_REPO

Set up all necessary project files in the current working directory and set appropriate
Julia preferences for using the system-installed MPI and p4est libraries. Furthermore, this
will install all dependencies in a local Julia depot folder (unless changed with the
'--julia-depot' flag).

The following files/folders will be created (or updated, if the '-f' flag is given):
- Project.toml
- Manifest.toml
- LocalPreferences.toml
- julia-depot

Note: Optional arguments with a value can also be set by defining the corresponding
environment variable prefixed by 'LIBTRIXI_'. For example, to use a Julia executable not on
the PATH, run this script as 'LIBTRIXI_JULIA_EXEC=julia-1.9.1 libtrixi-init-julia ...'.

positional arguments:
    PREFIX_OR_REPO  Path to install prefix of a libtrixi installation or, alternatively to
                    the root directory of a libtrixi clone. Note that the files at the
                    path must not be moved while libtrixi is used since that will break
                    library usage.

optional arguments:
    --help, -h      Show this help.

    --force, -f     By default, this tool does nothing if any of the files/folders to be
                    created already exist. With this option given on the command line, it
                    will continue anyways.

    --julia-depot JULIA_DEPOT
                    Path to the Julia depot to be used, i.e., the folder where all Julia
                    packages, Julia cache files etc. are stored. (default: 'julia-depot')

    --julia-exec JULIA_EXEC
                    Path to Julia executable. (default: 'julia')

    --mpi-library MPI_LIBRARY
                    Path to the MPI C shared library, i.e., something like
                    'path/to/libmpi.so'. If empty, Julia will try to figure out the path
                    automatically. (default: '')

    --p4est-library P4EST_LIBRARY
                    Path to the p4est shared library, i.e., something like
                    'path/to/libp4est.so'. If empty, Julia will try to figure out the path
                    automatically. (default: '')

    --sc-library SC_LIBRARY
                    Path to the sc shared library, i.e., something like
                    'path/to/libsc.so'. If omitted, it will be assumed to be located next
                    to the p4est library. (default: '$(dirname $P4EST_LIBRARY)/libsc.so')
```

sloede (approved):

sloede:
@bgeihe This PR looks good to me as it is. Once all checks pass, I would like to merge it if you don't mind, to continue working on #20. Would that be ok with you?

sloede:
@bgeihe If you don't object, I will just go ahead and merge this today, such that I can continue later. If there's a problem, we can of course always go back - also, the main change here is to add Trixi.jl and OrdinaryDiffEq.jl as dependencies, and this is easily undone.

PR #30, opened 2023-06-25
Use system HDF5 library

sloede:
Otherwise we get weird startup errors when calling `MPI.Init()` from Trixi.jl, since HDF5.jl pulls in OpenMPI from OpenMPI_jll.jl (via HDF5_jll.jl), which causes all kinds of errors upon initializing MPI.


sloede:
For some reason, tests are not reported as finished, although they have clearly passed and coverage information is also available. Merging this anyways...
image

PR #33, opened 2023-06-26
Add ability to control debug output via `LIBTRIXI_DEBUG`

### LINKED ISSUE #31: Add environment-variable controlled debug output

sloede:
Resolves #31.


bgeihe:
Nice!

Will the coverage checks realize that the output branches are taken once the CI uses the new environment variable?

bgeihe (dismissed):

sloede:
> Nice!
> 
> Will the coverage checks realize that the output branches are taken once the CI uses the new environment variable?

Good point! They do not, and I think we should enable those 👍 let's see if 0bc4f53 fixes it...

bgeihe (approved):

PR #34, opened 2023-06-26
Document troubleshooting

bgeihe:
Issues encountered during development of #20


docs/src/issues.md
```diff
@@ -0,0 +1,17 @@
+# Known issues
+
+## Missing cuda libraries
```
sloede, line 3: I feel like `Missing CUDA libraries` is not optimal for a person that does not even know that this issue is about CUDA. What about something like
```suggestion
## `dlopen failed` is triggered by OpenMPI
```
or something similar?

sloede (approved):

PR #39, opened 2023-06-28
Fix call of invokelatest for julia 1.8

### LINKED ISSUE #38: invokelatest not exported for julia 1.8
When using `julia` version 1.8 the following error occurs:
```
[...]
UndefVarError(var=:invokelatest)
[...]
```
Do we need to replace `invokelatest` by `Base.invokelatest`?

bgeihe:
Resolves #38

### DIFF
--- a/LibTrixi.jl/src/api_jl.jl
+++ b/LibTrixi.jl/src/api_jl.jl
@@ -4,6 +4,7 @@ function trixi_initialize_simulation_jl(filename)
 
     # Initialize simulation state
     # Note: we need `invokelatest` here since the function is dynamically upon `include`
+    # Note: `invokelatest` is not exported until Julia v1.9, thus we call it through `Base`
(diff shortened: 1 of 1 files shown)


LibTrixi.jl/src/api_jl.jl
```diff
@@ -4,7 +4,7 @@ function trixi_initialize_simulation_jl(filename)
 
     # Initialize simulation state
     # Note: we need `invokelatest` here since the function is dynamically upon `include`
-    simstate = invokelatest(Main.init_simstate)
+    simstate = Base.invokelatest(Main.init_simstate)
```
sloede, line 7:
```suggestion
    # Note: `invokelatest` is not exported until Julia v1.9, thus we call it through `Base`
    simstate = Base.invokelatest(Main.init_simstate)
```
Maybe as self-documentation for us?

sloede (approved):

PR #45, opened 2023-07-05
First steps towards accessing Trixi data

bgeihe:
Not intended for merging, rather for getting used to Trixi.


LibTrixi.jl/src/LibTrixi.jl
```diff
@@ -18,6 +18,10 @@ export trixi_is_finished,
 export trixi_step,
        trixi_step_cfptr,
        trixi_step_jl
+export trixi_ndims_cfptr
+export trixi_nelements_cfptr
+export trixi_nvariables_cfptr
+export trixi_get_cell_averages_cfptr
```
bgeihe, line 24: Is this sufficient? 
Do we actually need the export, or could we just write `LibTrixi.trixi_*`

LibTrixi.jl/src/api_c.jl
```diff
@@ -129,3 +129,64 @@ Base.@ccallable function trixi_step(simstate_handle::Cint)::Cvoid
 end
 
 trixi_step_cfptr() = @cfunction(trixi_step, Cvoid, (Cint,))
+
```
bgeihe, line 132: Brain can be switched off from here on...

LibTrixi.jl/src/api_jl.jl
```diff
@@ -49,3 +49,53 @@ function trixi_step_jl(simstate)
...
+    # convert C to julia
+    data = unsafe_wrap(Array, data_, nelements*nvariables)
+
+    u_ode = simstate.integrator.u
+    u = Trixi.wrap_array(u_ode, mesh, equations, solver, cache)
+
+    for (index, element) in enumerate(Trixi.eachelement(solver, cache))
+
+        # temporary storage for mean value on current element for all variables
+        u_mean = zero(Trixi.get_node_vars(u, equations, solver, 1, 1, element))
```
bgeihe, line 83: This could be spared.

src/trixi.c
```diff
@@ -190,6 +196,69 @@ void trixi_finalize() {
 }
 
 
+/** Return number of spatial dimensions
```
bgeihe, line 199: \brain OFF

LibTrixi.jl/src/LibTrixi.jl
sloede: We do not necessarily need the export, you are right. One can always use the qualified version, with `LibTrixi.` prefixed.

However, the `export` also highlights that a given function is part of the public API. That is, it is a user-facing function and not an internal implementation detail. That's why I exported all the functions each time: the `fun`, the `fun_cfptr`, and the `fun_jl`, since they each make sense to be used by a user IMHO.

LibTrixi.jl/src/api_jl.jl
```diff
@@ -49,3 +49,53 @@ function trixi_step_jl(simstate)
...
+    nelements = Trixi.nelements(solver, cache)
+    nvariables = Trixi.nvariables(equations)
+
+    # convert C to julia
+    data = unsafe_wrap(Array, data_, nelements*nvariables)
+
+    u_ode = simstate.integrator.u
+    u = Trixi.wrap_array(u_ode, mesh, equations, solver, cache)
+
+    for (index, element) in enumerate(Trixi.eachelement(solver, cache))
```
sloede, line 80: I think `index` and `element` will always be the same here. `eachelement` just goes from 1 to `nelements`, and the same will happen for `index` (I guess)

LibTrixi.jl/src/LibTrixi.jl
```diff
@@ -18,6 +18,18 @@ export trixi_is_finished,
...
+       trixi_ndims_jl
+export trixi_nelements,
+       trixi_nelements_cfptr,
+       trixi_nelements_jl
+export trixi_nvariables,
+       trixi_nvariables_cfptr,
+       trixi_nvariables_jl
+export trixi_get_cell_averages,
+       trixi_get_cell_averages_cfptr,
+       trixi_get_cell_averages_jl
```
sloede, line 32:
```suggestion
export trixi_load_cell_averages,
       trixi_load_cell_averages_cfptr,
       trixi_load_cell_averages_jl
```
I like the idea from Julia to use `load`/`store` to underline that something "heavy" is transferred. `get`/`set` could also be used for just passing around pointers, while `load`/`store` tell me that this means that a non-negligible amount of memory is likely to be shifted around.

But this is really only a suggestion and depends a lot on personal preference, so proceed as you see fit

LibTrixi.jl/src/api_c.jl
```diff
@@ -129,3 +129,64 @@ Base.@ccallable function trixi_step(simstate_handle::Cint)::Cvoid
...
+
+"""
+    trixi_ndims(simstate_handle::Cint)::Cint
+
+Return number of spatial dimensions
```
sloede, line 137:
```suggestion
Return number of spatial dimensions.
```
For consistency

LibTrixi.jl/src/api_c.jl
```diff
@@ -129,3 +129,64 @@ Base.@ccallable function trixi_step(simstate_handle::Cint)::Cvoid
...
+    return trixi_ndims_jl(simstate)
+end
+
+trixi_ndims_cfptr() = @cfunction(trixi_ndims, Cint, (Cint,))
+
+
+"""
+    trixi_nelements(simstate_handle::Cint)::Cint
+
+Return number of elements (cells)
```
sloede, line 152:
```suggestion
Return number of elements (cells).
```

LibTrixi.jl/src/api_c.jl
```diff
@@ -129,3 +129,64 @@ Base.@ccallable function trixi_step(simstate_handle::Cint)::Cvoid
...
+    return trixi_nelements_jl(simstate)
+end
+
+trixi_nelements_cfptr() = @cfunction(trixi_nelements, Cint, (Cint,))
+
+
+"""
+    trixi_nvariables(simstate_handle::Cint)::Cint
+
+Return number of elements (cells)
```
sloede, line 167: Bad copy/paste?

LibTrixi.jl/src/api_c.jl
```diff
@@ -129,3 +129,64 @@ Base.@ccallable function trixi_step(simstate_handle::Cint)::Cvoid
...
+    return trixi_nvariables_jl(simstate)
+end
+
+trixi_nvariables_cfptr() = @cfunction(trixi_nvariables, Cint, (Cint,))
+
+
+"""
+    trixi_get_cell_averages(simstate_handle::Cint)::Cint
+
+Return number of elements (cells)
```
sloede, line 182: Bad copy/paste?

LibTrixi.jl/src/api_jl.jl
```diff
@@ -49,3 +49,53 @@ function trixi_step_jl(simstate)
...
+    return Trixi.nelements(solver, cache)
+end
+
+function trixi_nvariables_jl(simstate)
+    _, equations, _, api_jl = Trixi.mesh_equations_solver_cache(simstate.semi)
+    return Trixi.nvariables(equations)
+end
+
+function trixi_get_cell_averages_jl(data_, simstate)
+
```
sloede, line 69:
```suggestion
```
No blank line at start of blocks

LibTrixi.jl/src/api_jl.jl
sloede: *bump*

LibTrixi.jl/src/api_jl.jl
```diff
@@ -49,3 +49,53 @@ function trixi_step_jl(simstate)
 
     return nothing
 end
+
+function trixi_ndims_jl(simstate)
+    mesh, _, _, _ = Trixi.mesh_equations_solver_cache(simstate.semi)
```
sloede, line 54: Instead of using `Trixi.xxx` here and elsewhere, you could just add all required symbols to the `using Trixi: ..., xxx` statement in `src/LibTrixi.jl`. But I'll leave this up to you (I'd just prefer something consistent, such that one does not have to think - either *always* add them to the `using` statement, or *never* add them and use the qualified access everywhere, as you do here)

LibTrixi.jl/src/api_jl.jl
```diff
@@ -49,3 +49,53 @@ function trixi_step_jl(simstate)
...
+    return ndims(mesh)
```
sloede, line 55: For example, this will probably not work, but would have to be
```suggestion
    return Trixi.ndims(mesh)
```
since I don't see it in the `using` statement

LibTrixi.jl/src/api_jl.jl
sloede, line 83: You could initialize `u_mean` outside of the loop and just reset it to zero at the beginning of each iteration

LibTrixi.jl/src/api_jl.jl
```diff
@@ -49,3 +49,53 @@ function trixi_step_jl(simstate)
...
+    u_ode = simstate.integrator.u
+    u = Trixi.wrap_array(u_ode, mesh, equations, solver, cache)
+
+    for (index, element) in enumerate(Trixi.eachelement(solver, cache))
+
+        # temporary storage for mean value on current element for all variables
+        u_mean = zero(Trixi.get_node_vars(u, equations, solver, 1, 1, element))
+
+        # compute mean value using nodal dg values and quadrature
+        for j in Trixi.eachnode(solver), i in Trixi.eachnode(solver)
```
sloede, line 86: This is not dimension agnostic. I think we should at least support 2D and 3D out of the box (but really also 1D).

You don't need to change it here, but maybe add a TODO note to the code and create an issue for it?

LibTrixi.jl/src/api_jl.jl
sloede, line 80: Instead of writing the integration manually, you could use the `integrate` function from Trixi.jl:
https://github.com/trixi-framework/Trixi.jl/blob/0816ed0b62679bcd656dc38bad68034843632ba1/src/callbacks_step/analysis_dg2d.jl#L205-L214
That would probably also make this dimension agnostic.

LibTrixi.jl/src/api_jl.jl
```diff
@@ -49,3 +49,53 @@ function trixi_step_jl(simstate)
...
+            u_node_prim = Trixi.cons2prim(Trixi.get_node_vars(u, equations, solver, i, j, element), equations)
```
sloede, line 87: It should at least be documented in the API that this does not get the mean values of the states, but the mean values of the primitive variables

examples/trixi_controller_data.c
```diff
@@ -0,0 +1,65 @@
...
+    double* data = malloc( sizeof(double) * nelements * nvariables );
+
+    // get averaged cell values for each variable
+    trixi_get_cell_averages(data, handle);
+
+    // compute temperature
+    const double gas_constant = 0.287;
+
+    for (int i = 0; i < nelements; ++i) {
+
```
sloede, line 52:
```suggestion
```

LibTrixi.jl/src/api_jl.jl
```diff
@@ -49,3 +49,53 @@ function trixi_step_jl(simstate)
...
+            u_mean += u_node_prim * solver.basis.weights[i] * solver.basis.weights[j]
+        end
+
+        # normalize to unit element
+        u_mean = u_mean / 2^ndims(mesh)
+
+        # copy to provided array
+        for ivar = 0:nvariables-1
+            data[index + ivar * nelements] = u_mean[ivar+1]
```
sloede, line 96: It seems like you store the data as
```
v1 v1 v1 .... v2 v2 v2 ... v3 v3 v3 ...
```
this should imho also be documented somewhere

examples/trixi_controller_data.f90
```diff
@@ -0,0 +1,74 @@
...
+  ! get number of variables
+  nvariables = trixi_nvariables( handle );
+  write(*, '(a,i6)') "*** Trixi controller ***   nvariables ", nelements
+  write(*, '(a)') ""
+
+  ! allocate memory
+  allocate ( u(0:nelements*nvariables) )
+
+  ! get averaged cell values for each variable
+  call trixi_get_cell_averages(u,handle);
```
sloede, line 60:
```suggestion
  call trixi_get_cell_averages(u, handle);
```

src/trixi.f90
```diff
@@ -114,6 +114,60 @@ subroutine trixi_finalize_simulation(handle) bind(c)
...
+    !!
+    !! @brief Return cell averaged values
+    !!
+    !! @param[in]  handle  simulation handle
+    !! @param[out] data    cell averaged values for all cells and all variables
+    !!
+    !! @see @ref trixi_get_cell_averages_api_c "trixi_get_cell_averages (C API)"
+    subroutine trixi_get_cell_averages(data, handle) bind(c)
+      use, intrinsic :: iso_c_binding, only: c_int, c_double
+      real(c_double), intent(in) :: data(*)
```
sloede, line 167:
```suggestion
      real(c_double), dimension(*), intent(in) :: data
```
For consistency with the interfaces above?

sloede (changes requested):
Overall looks quite nice to me! I've done a "somewhat" in-depth review, so there a lots of small details I noticed. But I'll leave it up to you if you want to merge this fast now and clean up later, or do it thoroughly right now.

LibTrixi.jl/src/api_jl.jl
bgeihe: done now

LibTrixi.jl/src/api_jl.jl
bgeihe: Arg, that's true.

LibTrixi.jl/src/api_jl.jl
bgeihe: Actually this is the place from where I tried to understand how to do it.
However, I need an integration per element, and not on the whole domain.

LibTrixi.jl/src/api_jl.jl
bgeihe: I thought `ndims` was just some basic julia statement.

LibTrixi.jl/src/api_jl.jl
sloede: Right, my bad...

LibTrixi.jl/src/api_jl.jl
sloede: That's true. Maybe you're right and the method is automatically added when doing `using Trixi: Trixi`, but I would try to test this if I were you.

sloede:
In general I noticed that you do not seem to test the new controllers... are they even built in the tests?

LibTrixi.jl/src/api_c.jl
```diff
@@ -177,16 +177,21 @@ trixi_nvariables_cfptr() = @cfunction(trixi_nvariables, Cint, (Cint,))
...
-    trixi_get_cell_averages(simstate_handle::Cint)::Cint
+    trixi_load_cell_averages(simstate_handle::Cint)::Cint
 
-Return number of elements (cells)
+Return cell averaged values
+
+Cell averaged values for each cell and each primitive variable are stored in a contiguous
+array, where cell values for the first variable appear first and values for the other
+variables subsequently.
+The given array has to be of correct size and memory has to be allocated beforehand.
```
sloede, line 187:
```suggestion
variables subsequently (structure-of-arrays layout).

The given array has to be of correct size and memory has to be allocated beforehand.
```

sloede: If any docstrings are changed here, they should probably also be changed in the C API docs.

LibTrixi.jl/src/api_jl.jl
```diff
@@ -49,3 +49,60 @@ function trixi_step_jl(simstate)
...
+    # all permutations of nodes indices for arbitrary dimension
+    node_cis = CartesianIndices(ntuple(i -> n_nodes, n_dims))
+
+    # temporary storage for mean value on current element for all variables
+    u_mean = get_node_vars(u, equations, solver, node_cis[1], 1)
+
+    for element in eachelement(solver, cache)
+
+        # compute mean value using nodal dg values and quadrature
+        u_mean = zero(u_mean)
```
bgeihe, line 87: Here, `u_mean` should be reset to zero.
Does this allocate a new `SVector` every time?
Even if this is true, something like `u_mean .= zero(eltype(..))` does not work. Because `SVector` is immutable?

LibTrixi.jl/src/api_jl.jl
```diff
@@ -49,3 +49,60 @@ function trixi_step_jl(simstate)
...
+        for node_ci in node_cis
```
bgeihe, line 88: This should now be dimension agnostic.
Later I realized that something very similar happens in `Trixi.trixi2tec`.

examples/trixi_controller_data.f90
```diff
@@ -0,0 +1,78 @@
...
+  write(*, '(a,i6)') "*** Trixi controller ***   nelements ", nelements
+  write(*, '(a)') ""
+
+  ! get number of variables
+  nvariables = trixi_nvariables( handle );
+  write(*, '(a,i6)') "*** Trixi controller ***   nvariables ", nvariables
+  write(*, '(a)') ""
+
+  ! allocate memory
+  allocate ( data(nelements*nvariables) )
```
bgeihe, line 58: Is this correct?
1-based and of length `nelements*nvariables`?

examples/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 set ( EXAMPLES
       simple_trixi_controller.c
-      simple_trixi_controller.f90 )
+      simple_trixi_controller.f90
+      trixi_controller_data.c
+      trixi_controller_data.f90 )
```
bgeihe, line 5: What do you think? Should the data access be demonstrated in separate examples like those, or integrated into the existing ones?

LibTrixi.jl/src/api_jl.jl
sloede: `SVector`s live on the stack and thus are not allocated. Thus this should be OK.

LibTrixi.jl/src/api_jl.jl
sloede: Sounds good!

examples/trixi_controller_data.f90
sloede: Ha, good question. But I think yes, this looks good to me!

examples/CMakeLists.txt
sloede: I think for now I'd leave it as it is until we figure out a different approach is better/more useful.

sloede (approved):
LGTM!

sloede:
Coverage is down because the C API functions are not used. I think you should call the new test program in our CI testing to fix this, e.g., by adding corresponding calls [here](https://github.com/trixi-framework/libtrixi/blob/d71c53c30c0c8e971ef993fc13a84d5d42cb6d23/.github/workflows/ci.yml#L154-L161) and [here](https://github.com/trixi-framework/libtrixi/blob/d71c53c30c0c8e971ef993fc13a84d5d42cb6d23/.github/workflows/ci.yml#L169-L176).

sloede (approved):

PR #48, opened 2023-07-06
Generalize examples cmake file

bgeihe:
This should make it easier to quickly add new examples for (local) testing.


sloede (approved):
LGTM, thanks!

PR #50, opened 2023-07-07
Simplify usage by allow to omit the JULIA_DEPOT_PATH

sloede:
Before, one always had to start programs with `JULIA_DEPOT_PATH=path myprog` when running a program `myprog` that links to `libtrixi`. With this PR, setting `JULIA_DEPOT_PATH` becomes unnecessary as long as the default depot location used by by `utils/libtrixi-init-julia` is left untouched.

This should help to avoid user errors in case they forget to set the Julia depot path appropriately. If a `JULIA_DEPOT_PATH` environment variable is set, it will honored, thus the user can still provide non-standard locations via the environment. Overriding the default is also possible by calling `trixi_initialize` with a second, non-null argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,12 +84,13 @@ cd libtrixi-julia
     --p4est-library <p4est_install_directory>/lib/libp4est.so
     <install_directory>
 ```
-
 Use `libtrixi-init-julia -h` to get help.
-When running a program that uses libtrixi, make sure to set up the `JULIA_DEPOT_PATH`
-environment variable to point to the `<julia-depot>` folder reported. In your code, pass
-the path to the `libtrixi-julia` directory to `trixi_initialize`, see the code of the
-examples.
+
+In your code, pass the path to the `libtrixi-julia` directory to `trixi_initialize`,
+see the code of the examples. If you did not modify the default value for the Julia depot
+when calling `libtrixi-init-julia`, libtrixi will find it automatically.
+Otherwise, when running a program that uses libtrixi, you need to make sure to set the
+`JULIA_DEPOT_PATH` environment variable to point to the `<julia-depot>` folder reported. 
 
 ### Testing
 
@@ -97,7 +98,6 @@ Go to some directory from where you want to run a Trixi simulation.
 
 ```shell
 LIBTRIXI_DEBUG=all \
-JULIA_DEPOT_PATH=<julia-depot> \
     <install_directory>/bin/simple_trixi_controller_c \
     <libtrixi-julia_directory> \
     <install_directory>/share/libtrixi/LibTrixi.jl/examples/libelixir_tree1d_dgsem_advection_basic.jl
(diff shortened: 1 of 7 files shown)


sloede:
@KerstinHartung This should also make it easier for MESSy users, since one cannot forget to set the depot path anymore if one sticks to the default values...

README.md
```diff
@@ -84,20 +84,20 @@ cd libtrixi-julia
...
-When running a program that uses libtrixi, make sure to set up the `JULIA_DEPOT_PATH`
-environment variable to point to the `<julia-depot>` folder reported. In your code, pass
-the path to the `libtrixi-julia` directory to `trixi_initialize`, see the code of the
-examples.
+
+In your code, pass the path to the `libtrixi-julia` directory to `trixi_initialize`,
+see the code of the examples. If you did not modify the default value for the Julia depot
+when calling `libtrixi-init-julia`, libtrixi will find it automatically.
+Otherwise, when running a program that uses libtrixi, you need to make sure to set the
+`JULIA_DEPOT_PATH` environment variable to point to the `<julia-depot>` folder reported.
```
bgeihe, line 93: Would you want to mention the other option, by supplying the depot path to `trixi_initialize`?

bgeihe (approved):
This a fantastic idea! I will definitely make using libtrixi more easy!

README.md
sloede: No, this should imho go into the API docs (whenever they come :-))

docs/src/index.md
```diff
@@ -86,20 +86,20 @@ cd libtrixi-julia
...
-When running a program that uses libtrixi, make sure to set up the `JULIA_DEPOT_PATH`
-environment variable to point to the `<julia-depot>` folder reported. In your code, pass
-the path to the `libtrixi-julia` directory to `trixi_initialize`, see the code of the
-examples.
+
+In your code, pass the path to the `libtrixi-julia` directory to `trixi_initialize`,
+see the code of the examples. If you did not modify the default value for the Julia depot
+when calling `libtrixi-init-julia`, libtrixi will find it automatically.
+Otherwise, when running a program that uses libtrixi, you need to make sure to set the
+`JULIA_DEPOT_PATH` environment variable to point to the `<julia-depot>` folder reported.
```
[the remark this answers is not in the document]
sloede: As above. I think that the README should describe the default use case and everything else is left for the detailed docs.

src/trixi.f90
```diff
@@ -2,9 +2,10 @@ module LibTrixi
   implicit none
 
   interface
-    subroutine trixi_initialize_c(project_directory) bind(c, name='trixi_initialize')
-      use, intrinsic :: iso_c_binding, only: c_char
+    subroutine trixi_initialize_c(project_directory, depot_path) bind(c, name='trixi_initialize')
+      use, intrinsic :: iso_c_binding, only: c_char, c_int
```
sloede, line 6:
```suggestion
      use, intrinsic :: iso_c_binding, only: c_char
```

PR #52, opened 2023-07-10
Generate C/Fortran API docs with Doxygen

sloede:
To generate the docs, go to `docs/doxygen` and run `doxygen` without any arguments. You can then view the result by opening the file `docs/doxygen/build/html/index.html` in your browswer.

This is just a very early draft after playing around some (too much) time with Doxygen 🙄 One particular challenge is the fact that we have to document APIs in two languages (actually three, if you include Julia), and that some function names are identical. I believe that this similarity is the reason why, e.g., the documentation for the Fortran interface might not show up properly (or maybe it just doesn't support Fortran interfaces).

Instead of using Documenter.jl for the main docs and Doxygen for the C/Fortran API, we could also consider to use Doxygen only for the C part and using something like [FORD](https://github.com/Fortran-FOSS-Programmers/ford) for the Fortran part.

Anyhow, this is as far as I'm going to take it with docs. @bgeihe feel free to adopt this PR, add to it, or just close it again if you don't like it - I just didn't want to get the time I put into this go to waste entirely :-)

The following things one would have to do if willing to make this PR work:
- [x] Figure out how to properly document the Fortran API
- [x] Auto-generate the docs as part of the call to Documenter.jl and put the result in a subdirectory
- [x] Link to C/Fortran API from main docs


sloede:
Preview the docs here: https://trixi-framework.github.io/libtrixi/previews/PR52

Go to "Reference" ➡️ "C/Fortran", where you will find a link to the doxygen-generated docs: https://trixi-framework.github.io/libtrixi/c-fortran-api

.github/workflows/Documenter.yml
```diff
@@ -45,3 +45,24 @@ jobs:
...
+        uses: actions/checkout@v3
+      - name: Install doxygen
+        run: |
+          sudo apt-get install -y doxygen
+      - name: Build doxygen docs
+        run: |
+          cd docs/doxygen
+          doxygen
+      - name: Deploy doxygen docs (only on main)
+        # if: ${{ github.ref == 'refs/heads/main' }} # TODO remove comment before merge!
```
sloede, line 63:
```suggestion
        if: ${{ github.ref == 'refs/heads/main' }}
```
Needs to be applied before merging the PR, otherwise future PR builds will overwrite the main docs

docs/src/reference-julia.md
```diff
@@ -0,0 +1,19 @@
+# Julia API
+
+This page documents the Julia part of libtrixi, which is implemented in the LibTrixi.jl
+package. LibTrixi.jl provides Julia-based wrappers around Trixi.jl, making simulations
+controllable through a defined API.
+
+!!! note "Internal and development use only"
+    The Julia API provided by LibTrixi.jl is only provided for internal use by libtrixi and
+    to facilitate easier development and debugging of new library features. It is *not*
+    intended to be used by Julia developers: They should directly utilize Trixi.jl to
```
bgeihe, line 10: This refers to users, who want to use Trixi but need not use the C / Fortran interface?

bgeihe:
I would like to check the output of [moxygen](https://github.com/sourcey/moxygen), but I can do this in a separate as well.

utils/libtrixi-init-julia
```diff
@@ -312,12 +312,12 @@ echo
...
+echo "    $PWD"
 echo
-echo "When calling libtrixi's 'trixi_initialize' function, you need to provide"
+echo "as the 'project_directory'. The environment variable"
 echo
-echo "    $PWD"
+echo "    JULIA_DEPOT_PATH=\"$PWD/$julia_depot\""
 echo
-echo "as the 'project_directory'."
+echo "will then automatically be set. It may be overriden manually."
```
bgeihe, line 323:
```suggestion
echo "will then automatically be set. It may be overridden manually."
```

docs/src/reference-julia.md
sloede: Yes exactly. Do you think it is unclear? Feel free to change/improve it 👍

.github/workflows/Documenter.yml
sloede: Not needed anymore.

sloede:
> I would like to check the output of [moxygen](https://github.com/sourcey/moxygen), but I can do this in a separate PR as well.

The current state of the PR is that everything _seems_ to work as it should, including the placement of the C/Fortran API into the proper folder such that it respects the versioning of Documenter.jl.

I thus recommend we go ahead and merge this PR in its current state (after a review by you, of course). All improvements can then be iterated on in a new PR.

docs/src/reference-julia.md
bgeihe: Thinking about it again, it seems clear.

bgeihe (approved):

PR #54, opened 2023-07-14
Work around MPI_set_attr error

bgeihe:
Hotfix, that hopefully can be removed again later.

https://github.com/JuliaParallel/MPI.jl/issues/746

### DIFF
--- a/src/trixi.c
+++ b/src/trixi.c
@@ -111,6 +111,18 @@ void trixi_initialize(const char * project_directory, const char * depot_path) {
         print_and_die("null pointer", LOC);
       }
     }
+
+    // TODO: Workaround for issue with MPI.jl: A required MPI key value is not initialized
+    //       in case MPI has already been initialized before. With libtrixi MPI could have
+    //       been initialized by an external application. Check if MPI is initialized but
+    //       the key value not. In this case run the init hooks manually.
+    //       This would better be fixed using some functionality of MPI.jl
+    const char * call_MPI_run_init_hook =
+      "using MPI;\n"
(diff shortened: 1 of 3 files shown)


src/trixi.c
```diff
@@ -111,6 +111,18 @@ void trixi_initialize(const char * project_directory, const char * depot_path) {
...
     }
+
+    // TODO: Workaround for issue with MPI.jl: A required MPI key value is not initialized
+    //       in case MPI has already been initialized before. With libtrixi MPI could have
+    //       been initialized by an external application. Check if MPI is initialized but
+    //       the key value not. In this case run the init hooks manually.
+    //       This would better be fixed using some functionality of MPI.jl
+    const char * call_MPI_run_init_hook =
+      "using MPI;\n"
+      "if MPI.Initialized() && MPI.JULIA_TYPE_PTR_ATTR[]==0;\n"
```
sloede, line 122: Since we are already tinkering with MPI.jl's internals here, why not use the internal variable that is set in `run_init_hooks`,

https://github.com/JuliaParallel/MPI.jl/blob/2a2bd68a6dc53660108bbc16690e4972de6318d5/src/MPI.jl#L55

to check if we need to run it again? That is, something like
```julia
      "if MPI.Initialized() && !MPI._finished_loading[];\n"
```
I am just concerned that this check might fail if `JULIA_TYPE_PTR_ATTR` is left == 0 by some MPI implementation.

sloede (commented):
Thanks a lot for providing this hot fix! Maybe you can also create an issue that lets us track this such that we cannot forget to fix it once there is a canonical solution available from MPI.jl.

PR #64, opened 2023-08-01
Use imported target for MPI

bgeihe:
When mpi does not reside in a default location, cmake's `FindCMake` module might still find it and will add `-lmpi` as a library to link. However, if neither `LD_LIBRARY_PATH` nor some other mechanism includes the location, the actual linking fails. `MPI_C_LINK_FLAGS` contains the required directory, but is currently not used. 

Since we meanwhile require cmake version 3.9, we can use the imported target, which magically solves the issue.


sloede:
Also, if this works, we should probably also update the cmake file in the examples folder.

sloede (commented):
LGTM!

PR #65, opened 2023-08-02
Canonical fix for MPI attr error

bgeihe:
As worked out here:
https://github.com/JuliaParallel/MPI.jl/issues/746

### DIFF
--- a/LibTrixi.jl/Project.toml
+++ b/LibTrixi.jl/Project.toml
@@ -6,6 +6,7 @@ version = "0.1.0"
 [deps]
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 Trixi = "a7f1ee26-1774-49b1-8366-f1abc58fbfcb"
+MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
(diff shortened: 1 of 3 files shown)


LibTrixi.jl/Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.1.0"
 [deps]
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 Trixi = "a7f1ee26-1774-49b1-8366-f1abc58fbfcb"
+MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
 
 [compat]
 julia = "1.8"
```
sloede, line 12:
```suggestion
julia = "1.8"
MPI = "0.20.13"
```

sloede (changes requested):
LGTM once the missing compat entry is added

LibTrixi.jl/Project.toml
bgeihe: Thank you! Missed it. Is it a general rule to always add packages to both `[deps]` and `[compat]`?

Issue #66, opened 2023-08-03
Add `trixi_version_julia` to provide info on Julia package versions


sloede:
Right now, `trixi_version` prints out the version of libtrixi, e.g., `0.1.0-dev`. I think it would be helpful for debugging purposes if we had an API function `trixi_version_julia` that outputs additional information about the underlying Julia packages currently used and Julia itself, e.g., something like this:
```
LibTrixi v0.1.0-pre
Trixi v0.5.34
OrdinaryDiffEq v5.26.3
julia v1.9.1
```

This could then also be automatically shown if `LIBTRIXI_DEBUG` is enabled.

bgeihe:
Sounds good!

I assume there is some nice juliaish way to accomplish this? Using `Pkg`? I would happily figure this out, unless you already know how to do it.

sloede:
> I assume there is some nice juliaish way to accomplish this? Using `Pkg`? I would happily figure this out, unless you already know how to do it.

Unfortunately, I don't. Getting the Julia version is trivial, but getting specific package versions from `Pkg` is not doable in a very elegant way, IIRC. But it would be great if you can take a stab at this.

bgeihe:
1. Currently we output the status of the julia project
  ```
  [f67ccb44] HDF5 v0.16.15
  [7e097bd5] LibTrixi v0.1.0 `../libtrixi_main/share/libtrixi/LibTrixi.jl`
  [da04e1cc] MPI v0.20.13
  [3da0fdf6] MPIPreferences v0.1.9
  [1dea7af3] OrdinaryDiffEq v6.53.4
  [7d669430] P4est v0.4.7
  [21216c6a] Preferences v1.4.0
  [d0cc0030] T8code v0.4.1
  [a7f1ee26] Trixi v0.5.37
  [cf7118a7] UUIDs
  ```
  I assume this is not sufficient as we are not able to access the actual data, but can merely print it?

2. Something like this seems to work
  ```
  julia> using Pkg
  julia> pkg_manifest = Pkg.Operations.Context().env.manifest
  julia> pkg_manifest[findfirst(v -> v.name == "Trixi", pkg_manifest)]
  PackageEntry(
    name = Trixi
    version = 0.5.37
    tree_hash = 31fa40be33d1189add1eab34547b745dbd2ebb17
  )
  ```
It would allow to query the version of any package, given by its name, recorded in the manifest.

sloede:
[...]

PR #68, opened 2023-08-06
Proof of concept for testing the C-julia-interface

bgeihe:
Not for merge, rather for discussion.

My overall idea was to call the C-API functions from julia (i.e. a loop julia -> C -> julia). One can then compare to corresponding direct calls of the julia API functions.


test/test_interface_c.jl
```diff
@@ -0,0 +1,80 @@
...
+
+    # get path of shared object, relative to LibTrixi.jl
+    libtrixi_so_path = joinpath(dirname(pathof(LibTrixi)), "../../../../lib/")
+    push!(Libdl.DL_LOAD_PATH, libtrixi_so_path)
+
+    # try to load the shared object
+    libtrixi_handle = dlopen("libtrixi")
+
+
+    # TODO: get version and compare to something (?)
```
bgeihe, line 17: tbd
also depending on #66

test/test_interface_c.jl
```diff
@@ -0,0 +1,80 @@
...
+        @test (@ccall "libtrixi".trixi_version_patch()::Int64) == 0
+    end
+
+
+    # initialize julia from C
+    @ccall "libtrixi".trixi_initialize(dirname(Base.active_project())::Cstring, C_NULL::Ptr{Cvoid})::Cvoid
+
+    # TODO: test the simulation state handling
+    #       therefore a libelexir is loaded
+    libelixir_path = joinpath(dirname(pathof(LibTrixi)), "../examples/libelixir_tree1d_dgsem_advection_basic.jl")
```
bgeihe, line 30: This means the test is based on just one example. (which is bad ?)

Alternatively one could construct a simstate mockup in a libelixir file. This could actually generate random pseudo simulation data (within reasonable bounds).
On the other hand, this approach would not test the real Trixi, OrdinaryDiffEq, ... workflow.

test/test_interface_c.jl
```diff
@@ -0,0 +1,80 @@
...
+
+    @testset verbose=true showtiming=true "Trixi simulation" begin
```
bgeihe, line 32: This testset is large, but I did not know how to create the simstate handle in one testset and keep it alive for the next set.

test/test_interface_c.jl
```diff
@@ -0,0 +1,80 @@
...
+
+        # finalize simulation from C
+        @ccall "libtrixi".trixi_finalize_simulation(handle_c::Cint)::Cvoid
+
+        # handle is now invalid
+        @test_throws ErrorException @ccall "libtrixi".trixi_is_finished(handle_c::Cint)::Cint
+    end
+
+    # finalize julia from C
+    # segmentation fault when executing jl_atexit_hook(0)
```
bgeihe, line 76: This crashed. 
Maybe it is not a good idea to call `jl_atexit_hook` while julia is still running?

sloede:
I am not fully convinced that testing our C and Fortran interfaces from Julia is the right move. For once, you already noticed that it skips and/or fails with some initialization/finalization routines. Furthermore, it might hide some issues about mismatching libraries etc. if Julia automatically picks up the correct libraries where an external program would not. Finally, this does not resolve testing the Fortran interface.

There might be good reasons for doing it this way that I am not seeing right now. But my initial feeling is that if we want to have robust testing (including parallel tests with MPI) for all 3 languages, we need to test each language interface separately.

bgeihe:
Two things motivated this approach:

1. I was happy with the julia testing facilities
2. It would have been easy to create mock-ups of LibTrixi simstates on-the-fly, which could then be used to compare to values returned via the interfaces.

Follow-up issue: #74

PR #71, opened 2023-08-12
Add `-Wall -Wextra -Werror` to compilation flags

### LINKED ISSUE #70: Add `-Wall -Wextra -Wpedantic -Werror` to default flags (if not yet done)
Following our discussion, we should make sure that we have at least `-Wall -Wextra -Werror` and possibly `-Wpedantic` enabled for both building libtrixi and our examples. This is an easy way to avoid weird errors that might be hard to debug.

sloede:
Resolves #70.

Note that we cannot add `-Wpedantic` since we convert the return value of `jl_unbox_voidpointer` to a function pointer when we want to call Julia functions from C. While this does not present an issue on x64 platforms, it is technically undefined behavior (C only explicitly allows object pointer to object pointer conversion, and function pointer to function pointer conversion). There does not seem to be a sane way around this at the moment, but since it is not presenting a problem either, I think it is ok for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,7 @@ target_link_libraries( ${PROJECT_NAME} PRIVATE ${JULIA_LIBRARY} MPI::MPI_C )
 
 # Set appropriate compile flags
 target_compile_options( ${PROJECT_NAME} PUBLIC "-fPIC" )
+target_compile_options( ${PROJECT_NAME} PRIVATE -Wall -Wextra -Werror)
 
 # Pass version information
 target_compile_definitions( ${PROJECT_NAME} PRIVATE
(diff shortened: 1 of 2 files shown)


bgeihe:
Fine with me (leaving out `-Wpedantic`)

CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ target_link_libraries( ${PROJECT_NAME} PRIVATE ${JULIA_LIBRARY} MPI::MPI_C )
 
 # Set appropriate compile flags
 target_compile_options( ${PROJECT_NAME} PUBLIC "-fPIC" )
+target_compile_options( ${PROJECT_NAME} PRIVATE -Wall -Wextra -Werror)
```
bgeihe, line 77: This will add the flags regardless of release or debug build type, right? 
But this is reasonable anyway.

sloede: Correct. But why wouldn't we 😬?

Issue #67, opened 2023-08-05
Add Julia tests for LibTrixi.jl 


sloede:
Right now, only the C Interface is really tested. If we want LibTrixi.jl to be considered also as a standalone package, we should add regular Julia tests as well.

bgeihe:
Just a quick note: I started reading up on this and implemented some initial pieces in a branch (which is currently not up to date though)

bgeihe:
Similar to the current C / Fortran test we could run a libelexir as explained in README.md

Do we want more, using `Test.jl`?

bgeihe:
:bulb:

bgeihe:
**Summary**

bgeihe:
**Summary**
Following julia code lines are not covered:

- `LibTrixi.jl`: In `show_debug_output()`: other choices for `LIBTRIXI_DEBUG`
- `api_c.jl`: all the `*_cfptr() = @cfunction(*, Cvoid, (...))` are not detected
- `api_jl.jl`: OrdinaryDiffEq integrator failing on doing a step
- `simulationstate.jl`
  - `store_simstate`: maximum count exceed, getting already existing handle (both should be near impossible)
  - `delete_simstate`: this should be tested...

Issue #57, opened 2023-07-15
Add workflow for automatic release creation


sloede:
For Julia-based repos we can use JuliaRegistrator together with TagBot to automatically create new releases. For libtrixi, we ideally want something similarly easy to use, maybe with the ability to create a release including a nice summary of merged PRs/closed issues since the last release upon pushing a tag with a pattern of `v#.#.#`.

Thus, we might want to have a workflow that uses one action to create a release summary/changelog (e.g., https://github.com/mikepenz/release-changelog-builder-action), and then passes it own to another action that creates the actual release (e.g., https://github.com/softprops/action-gh-release).

sloede:
With TagBot now working properly and the release process being [documented](https://trixi-framework.github.io/libtrixi/stable/developers/#Release-management), this can be closed.

PR #75, opened 2023-08-14
Initial implementation of `trixi_version_julia` package info

### LINKED ISSUE #62: Keep VERSION consistent to LibTrixi.jl/Project.toml
Addendum to #56 : Do we need and can we make sure that the versions in file `VERSION`  and `LibTrixi.jl/Project.toml` agree?

### LINKED ISSUE #66: Add `trixi_version_julia` to provide info on Julia package versions

bgeihe:
Meant to eventually resolve #66 
Also resolves #62


LibTrixi.jl/src/LibTrixi.jl
```diff
@@ -52,6 +66,22 @@ function show_debug_output()
     end
 end
 
+
+function assemble_version_info(;filter_expr = x -> x)
```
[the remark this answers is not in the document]
bgeihe: ìdentity` exists as a function? Awesome...

LibTrixi.jl/src/api_jl.jl
```diff
@@ -106,3 +106,11 @@ function trixi_load_cell_averages_jl(data, simstate)
...
     return nothing
 end
+
+function trixi_version_julia_jl()
+    return _version_info
+end
+
+function trixi_version_julia_extended_jl()
+    return _version_info_extended
+end
```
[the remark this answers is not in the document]
bgeihe: Correct, there are not used. They should be tested in the course of #67 .

And correct, that is a bug.

src/trixi.c
```diff
@@ -20,42 +20,34 @@ static int show_debug_output();
...
-  [TRIXI_FTPR_CALCULATE_DT]          = "trixi_calculate_dt_cfptr",
-  [TRIXI_FTPR_IS_FINISHED]           = "trixi_is_finished_cfptr",
-  [TRIXI_FTPR_STEP]                  = "trixi_step_cfptr",
-  [TRIXI_FTPR_FINALIZE_SIMULATION]   = "trixi_finalize_simulation_cfptr",
-  [TRIXI_FTPR_NDIMS]                 = "trixi_ndims_cfptr",
-  [TRIXI_FTPR_NELEMENTS]             = "trixi_nelements_cfptr",
-  [TRIXI_FTPR_NVARIABLES]            = "trixi_nvariables_cfptr",
-  [TRIXI_FTPR_LOAD_CELL_AVERAGES]    = "trixi_load_cell_averages_cfptr",
-};
-
```
bgeihe, line 53: For once I moved the definition of the function pointer arrays to the header file, s.t. they become available in multiple c-files.
It allows to more clearly organize the C-API implementation. On the other hand the definition and declaration of the array get cumbersome (need for `extern`, ..., see below).

sloede: The use of `extern` is canonical C and only required once in the header file, so that's ok imho.

However, what I wouldn't do is to put them in `trixi.h`. `trixi.h` is our externally visible include file and should only contain definitions of interest for an external user. OTOH, our function pointers are for internal use only. Thus, I would move the enum and the two global variable declarations to a new file, e.g., `function_pointers.h`, and the corresponding variable definitions to `function_pointers.c`.

sloede: I also noted that we still prefix all file names with `trixi_`. While this is technically ok, we only really need `trixi.h` to have "trixi" in its name. Since it seems like we are now spreading the code to more files now, we could also consider to rename the files after their purpose, e.g., `api.c` (instead of `trixi.c`) and plain `version.c` or something similar/better

bgeihe: Good points!

Maybe I revert the spreading and do a dedicated restructuring PR?

sloede: Up to you. You could also try to finish this PR quickly as-is, and then create a new PR for cleanup.

sloede:
> Meant to eventually resolve #66

Does it resolve the issue or not? Since you use the keyword "resolve", merging this PR will auto-close #66 as well. If this is the intent, great, if not, you should change your description 🙂

bgeihe:
> > Meant to eventually resolve #66
> 
> Does it resolve the issue or not?

I'd say it actually does.

It also resolves #62, I think.

However #57 remains open.

sloede:
> I'd say it actually does.
> 
> It also resolves #62, I think.

Then you should change the PR description to something like
```
Resolves #62, resolves #66.
```

bgeihe:
Looking at the diff again, I would really hat to remove the nice version parsing form `CMakeLists.txt`.
Couldn't we just parse `LibTrixi.jl/Project.toml`?

src/trixi.c
```diff
@@ -119,6 +127,12 @@ void trixi_initialize(const char * project_directory, const char * depot_path) {
         print_and_die("null pointer", LOC);
       }
     }
+
+    // Show version info
+    printf("\nLibTrixi %s\n\n", trixi_version());
```
sloede, line 132:
```suggestion
    printf("\nlibtrixi %s\n\n", trixi_version());
```
We style it as `libtrixi`, don't we?

LibTrixi.jl/src/api_jl.jl
```diff
@@ -106,3 +106,27 @@ function trixi_load_cell_averages_jl(data, simstate)
...
+    return VersionNumber(_version_libtrixi[]).major
+end
+
+function trixi_version_minor_jl()
+    return VersionNumber(_version_libtrixi[]).minor
+end
+
+function trixi_version_patch_jl()
+    return VersionNumber(_version_libtrixi[]).patch
+end
```
sloede, line 120: If we decide to keep those functions (see my other comment below), they should return `Int`s and not `UInt32` imho:
```suggestion
function trixi_version_major_jl()
    return Int(VersionNumber(_version_libtrixi[]).major)
end

function trixi_version_minor_jl()
    return Int(VersionNumber(_version_libtrixi[]).minor)
end

function trixi_version_patch_jl()
    return Int(VersionNumber(_version_libtrixi[]).patch)
end
```

sloede (changes requested):

LibTrixi.jl/src/LibTrixi.jl
```diff
@@ -61,6 +104,12 @@ function __init__()
     if show_debug_output()
         MPI.set_default_error_handler_return()
     end
+
+    # assemble packages information
+    _version_info[] = assemble_version_info(filter_expr = filter(p -> p.is_direct_dep))
+    _version_info_extended[] = assemble_version_info()
+    libtrixi_string = assemble_version_info(filter_expr = filter(p -> p.name == "LibTrixi"))
+    _version_libtrixi[] = split(split(libtrixi_string, "\n")[1], " ")[2]
```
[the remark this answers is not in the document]
bgeihe: Yap, not exactly brilliant.
Obviously you are very attentive!

CMakeLists.txt
```diff
@@ -1,19 +1,17 @@
 # Specify the minimum version (3.9 required for regex submatches).
 cmake_minimum_required ( VERSION 3.9 )
 
-## Get project version
-# Reconfigure if VERSION file has changed
-set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${CMAKE_SOURCE_DIR}/VERSION")
```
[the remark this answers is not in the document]
bgeihe: Yes, thanks!

CMakeLists.txt
```diff
@@ -1,19 +1,17 @@
...
-  set(LIBTRIXI_VERSION ${CMAKE_MATCH_0})
-  set(LIBTRIXI_VERSION_MAJOR ${CMAKE_MATCH_1})
-  set(LIBTRIXI_VERSION_MINOR ${CMAKE_MATCH_2})
-  set(LIBTRIXI_VERSION_PATCH ${CMAKE_MATCH_3})
+if (${libtrixi_project_toml} MATCHES "version = \"(([0-9]+)\\.([0-9]+)\\.([0-9]+)([+-][A-Za-z0-9_-]+)?)\"")
+  set(LIBTRIXI_VERSION ${CMAKE_MATCH_1})
+  set(LIBTRIXI_VERSION_MAJOR ${CMAKE_MATCH_2})
+  set(LIBTRIXI_VERSION_MINOR ${CMAKE_MATCH_3})
+  set(LIBTRIXI_VERSION_PATCH ${CMAKE_MATCH_4})
+  set(LIBTRIXI_VERSION_PRERELEASE ${CMAKE_MATCH_5})
```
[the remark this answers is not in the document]
bgeihe: Yes, it does. Thanks for the solution!

LibTrixi.jl/src/api_jl.jl
```diff
@@ -106,3 +106,27 @@ function trixi_load_cell_averages_jl(data, simstate)
...
+    return _version_libtrixi[]
+end
+
+function trixi_version_julia_jl()
+    return _version_info[]
+end
+
+function trixi_version_julia_extended_jl()
+    return _version_info_extended[]
+end
```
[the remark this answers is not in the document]
bgeihe: No no, I want to become more Julian...

sloede (changes requested):
Almost ready to merge, just one proposed change left.

PR #76, opened 2023-08-15
Create library with PackageCompiler.jl

sloede:
I played around with compiling libtrixi directly from the Julia sources (without having to use our C library layer). This is very much WIP and lots is missing, but I didn't want to lose the effort, so here we go...


bgeihe:
Enthused that you already got that started!
This is something hidden deeply in my notes.

sloede:
The decrease in coverage is imho a false negative. Other than that, I think this PR is ready for review and merge.

CMakeLists.txt
```diff
@@ -128,4 +129,5 @@ install( TARGETS ${PROJECT_NAME} )
 install( FILES ${CMAKE_Fortran_MODULE_DIRECTORY}/libtrixi.mod TYPE INCLUDE)
 install( FILES $<TARGET_OBJECTS:${PROJECT_NAME}_tls> TYPE LIB RENAME lib${PROJECT_NAME}_tls.o )
 install( DIRECTORY LibTrixi.jl DESTINATION share/libtrixi )
```
bgeihe, line 131: `LibTrixi.jl/lib` will be installed as well. Exclude it or just ignore?

README.md
```diff
@@ -252,6 +255,35 @@ that are linked with your main program. See `MakefileExternal` for an example of
...
 
+### Experimental support direct compilation of the Julia sources
+There is _experimental_ support for compiling the Julia sources in LibTrixi.jl to a shared
+library with a C interface. This is possible with the use of the Julia package
+[PackageCompiler.jl](https://github.com/JuliaLang/PackageCompiler.jl).
+
+To try this out, perform the following steps:
+1. Initialize the project directory `libtrixi-julia` using `libtrixi-init-julia` as
+   described above.
+2. Go to the `LibTrixi.jl/lib` directory and call `make`:
```
bgeihe, line 266:
```suggestion
2. Go to the `LibTrixi.jl/lib` directory in the repository root,
   make sure that `PROJECT_DIR` (defined in `Makefile`) points to your `libtrixi-julia` directory,
   and call `make`:
```

docs/src/index.md
```diff
@@ -252,6 +255,35 @@ that are linked with your main program. See `MakefileExternal` for an example of
...
 
+### Experimental support direct compilation of the Julia sources
+There is _experimental_ support for compiling the Julia sources in LibTrixi.jl to a shared
+library with a C interface. This is possible with the use of the Julia package
+[PackageCompiler.jl](https://github.com/JuliaLang/PackageCompiler.jl).
+
+To try this out, perform the following steps:
+1. Initialize the project directory `libtrixi-julia` using `libtrixi-init-julia` as
+   described above.
+2. Go to the `LibTrixi.jl/lib` directory and call `make`:
```
bgeihe, line 266:
```suggestion
2. Go to the `LibTrixi.jl/lib` directory in the repository root,
   make sure that `PROJECT_DIR` (defined in `Makefile`) point to your `libtrixi-julia` directory,
   and call `make`:
```

bgeihe (commented):
Impressive!

CMakeLists.txt
sloede: Good point. I think ignore for now, but maybe create an issue for it to make it cleaner, eg, move to an out-of-source build?

bgeihe: Fine with ignoring it.

But: Should triggering the PackageCompiler from the installed libtrixi be an option? (tests, e.g., are not installed and we only run them from the build directory)

sloede: No, I’d say that it’s either PackageCompiler or our C library. I would treat them as mutually exclusive, not a composition. So you either use our C library or the PC approach during installation. For that, we also need a new CMakeLists.txt (or better yet, a selector whether you want to use the C shim or PC) and then fix the installation part accordingly. What’s also missing right now is the ability to use the Fortran examples.

maybe you can take a stab at integrating the PC option into our CMake setup? 😬

bgeihe: Sounds like a nice idea. Moved to #124.

bgeihe (approved):

PR #78, opened 2023-08-15
Julia test for LibTrixi.jl 

### LINKED ISSUE #67: Add Julia tests for LibTrixi.jl

bgeihe:
Should at some point resolve #67


.github/workflows/ci.yml
```diff
@@ -165,6 +165,14 @@ jobs:
               ../build/examples/trixi_controller_data_f . libelixir_p4est2d_dgsem_euler_sedov.jl
         env:
           LIBTRIXI_DEBUG: all
+      - name: Run julia tests with coverage
+        if: ${{ matrix.test_type == 'coverage' }}
+        run: |
+          cd libtrixi-julia
+          JULIA_DEPOT_PATH=~/.julia \
+              julia --code-coverage=user --code-coverage=@.. --project=. ../test/runtests.jl
```
bgeihe, line 173: This generates `*.cov` files.

How to proceed? Merge them (as in `Trixi.jl`), export `lcov.info`, and merge it somehow with other one, generated from C-code?

.github/workflows/ci.yml
```diff
@@ -165,6 +165,21 @@ jobs:
               ../build/examples/trixi_controller_data_f . libelixir_p4est2d_dgsem_euler_sedov.jl
         env:
           LIBTRIXI_DEBUG: all
+      - name: Run julia tests with coverage
+        if: ${{ matrix.test_type == 'coverage' }}
+        uses: julia-actions/julia-runtest@v1
+        with:
+          coverage: true
+          project: ./libtrixi-julia
```
sloede, line 173: This should IMHO point to the LibTrixi.jl directory, not the runtime dir we created.

bgeihe: :pray:

.github/workflows/ci.yml
```diff
@@ -182,23 +197,27 @@ jobs:
...
           lcov --directory ../build  --capture --output-file lcov.info
+      - name: Process julia coverage data
+        uses: julia-actions/julia-processcoverage@v1
+        with:
+          directories: LibTrixi.jl/src
       - uses: codecov/codecov-action@v3
         if: ${{ matrix.test_type == 'coverage' }}
         with:
-          files: ./libtrixi-julia/lcov.info
+          files: ./libtrixi-julia/lcov.info ./lcov.info
```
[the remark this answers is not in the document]
bgeihe: You were right!
However we need to update coveralls, it seems.

LibTrixi.jl/Project.toml
```diff
@@ -25,3 +25,9 @@ PrecompileLowStorage = true
 PrecompileNoSpecialize = false
 PrecompileNonStiff = true
 PrecompileStiff = false
+
+[extras]
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[targets]
+test = ["Test"]
```
sloede, line 33:
```suggestion
```
Instead, create a `LibTrixi.jl/test/Project.toml` file with all relevant entries.

bgeihe:
**Summary**
Following julia code lines are not covered:

- `LibTrixi.jl`: In `show_debug_output()`: other choices for `LIBTRIXI_DEBUG`
- `api_c.jl`: all the `*_cfptr() = @cfunction(*, Cvoid, (...))` are not detected
- `api_jl.jl`: OrdinaryDiffEq integrator failing on doing a step
- `simulationstate.jl`
  - `store_simstate`: maximum count exceed, getting already existing handle (both should be near impossible)
  - `delete_simstate`: strange, this should be tested...

LibTrixi.jl/Project.toml
bgeihe: I read from the manual that either way is fine. And as I did not fully understand the second approach, I chose the first.

I now copied `Project.toml` to `test/` and added `test` there. Is this how it is supposed to be?

sloede:
> * `LibTrixi.jl`: In `show_debug_output()`: other choices for `LIBTRIXI_DEBUG`

Simple unit tests can help here.

> * `api_c.jl`: all the `*_cfptr() = @cfunction(*, Cvoid, (...))` are not detected

You can just call them in unit tests and verify that they return pointers that are not null.

> * `api_jl.jl`: OrdinaryDiffEq integrator failing on doing a step
> * `simulationstate.jl`

Too hard to test. Just ignore it.

>   * `store_simstate`: maximum count exceed, getting already existing handle (both should be near impossible)

Again, just ignore.

>   * `delete_simstate`: strange, this should be tested...

Can't really tell why, just keep digging I guess?

bgeihe:
Thanks for the hints!

bgeihe:
All addressed. Except `delete_simstste`, which only gets called from `trixi_finalize_simulation`, which however will check the handle before via `load_simstate`.

LibTrixi.jl/test/Project.toml
```diff
@@ -0,0 +1,23 @@
+[deps]
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
+Trixi = "a7f1ee26-1774-49b1-8366-f1abc58fbfcb"
+MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
sloede, line 5:
```suggestion
```

I think these are not needed (and neither are their compat entries). IIRC, you only need to add packages here that you use in `runtest.jl` (and its includes) - you do not need this for packages that are used by LibTrixi.jl, since they will already be installed automatically with LibTrixi.jl.

Please verify that I did not miss anything, but AFAICT, your `[deps]` should only contain the entry for `Test` and the `[compat]` section can be removed entirely.

sloede (changes requested):
Looks already quite good - with the remaining convo resolved, this can be merged.

Nicely done, upping the test coverage of the LibTrixi.jl folder from 0% to 97%! 🙌

LibTrixi.jl/test/Project.toml
bgeihe: Thank you for the explanation! To be honest, I did not fully understand the mechanism.

I tried to cleanup `test/Project.toml` now.
With only `Test` in it, I can run the tests successfully via
```
~/libtrixi/LibTrixi.jl $ JULIA_DEPOT_PATH=~/install/libtrixi-julia/julia-depot LIBTRIXI_DEBUG=all julia --project=./ test/runtests.jl
```
Via julia
```
~/libtrixi/LibTrixi.jl $ JULIA_DEPOT_PATH=~/install/libtrixi-julia/julia-depot LIBTRIXI_DEBUG=all julia --project=./ test/runtests.jl

julia> import Pkg

julia> Pkg.test()
```
however, it does not work. It is missing `Trixi` and tells me to `Pkg.add` it.

With `Test`, `Trixi`, and `OrdinaryDiffEq` in Project.toml it works again.

sloede: Can you please fix the Julia command for the second case? There seems to be a copy paste error...

bgeihe: Indeed!

sloede: I think if you just run `Pkg.test()`, the active environment should be the folder LibTrixi.jl. Is this the case? Also, what is the exact error? I don't understand why it would miss "Trixi", since `using Trixi` is never executed in the files located in `test/`

bgeihe: Another idea:
`test/runtests.jl` includes `test/test_interface.jl`, which calls `trixi_initialize_simulation(libelixir)`, which includes `libelixir_tree1d_dgsem_advection_basic.jl` (via `Base.include(Main, libelixir)`), which starts with
```
using LibTrixi
using Trixi
using OrdinaryDiffEq
``` 
Could this be the reason?

sloede: Hm. You might be right!

Is there a way to easily make the module configurable, i.e., such that it is `Main` by default (especially when called from C), but to allow one to use a different module from Julia? E.g., a second version of `trixi_initialize_simulation` with an additional first parameter `module`, and then connect them together similar to this:
```julia
function trixi_initialize_simulation(module, libelixir::String)
    ....
end
function trixi_initialize_simulation(libelixir::String)
    trixi_initialize_simulation(Main, libelixir)
end
```
?

bgeihe: Certainly! This could also be a parameter with default value, couldn't it?

But I do not get the point. Would I then used `Module=LibTrixi` when testing?

sloede: > Certainly! This could also be a parameter with default value, couldn't it?

Yes. However, this is how we do it in [Trixi.jl](https://github.com/trixi-framework/Trixi.jl/blob/e22b11e5859b5168e24dabb4cb34a6a35fce3e66/src/auxiliary/special_elixirs.jl#L38C24-L48) and how, I believe, is custom in other Julia base functions.

> Would I then used Module=LibTrixi when testing?

No, you'd use something like `@__MODULE__` which gives the currently active module. The only issue I can with this is that it will not allow us to fully test the C interface from within Julia anymore, since we cannot pass an object of type `Module` through the C interface. But maybe that's OK?

bgeihe: I'm afraid this is what I am currently doing. 
https://github.com/trixi-framework/libtrixi/blob/35485c1370437a69e7e51747447c6a6361355459/LibTrixi.jl/test/test_interface.jl#L23
But of course the test could be rewritten.

I assume `@__MODULE__` cannot just always be used?

bgeihe: The same as before
```
*** Base.include with module Main.TestInterface
libtrixi tests: Error During Test at /home/bene/libtrixi_main/LibTrixi.jl/test/runtests.jl:3
  Got exception outside of a @test
  LoadError: ArgumentError: Package Trixi not found in current path.
  - Run `import Pkg; Pkg.add("Trixi")` to install the Trixi package.
```

sloede: Can you also give me the full command you use to trigger this error with the current branch?

I don't think that evaluating this inside the module `LibTrixi` is the right choice, which is why I am hesitant...

bgeihe: Sure! I just pushed the experimental version.

```
~/libtrixi_main/LibTrixi.jl (bg/julia-test *+) $ LIBTRIXI_DEBUG=all julia --project=./
"]"
(LibTrixi) pkg> test
```

Don't get me wrong! I am not claiming that it would be right to do it in another way. I just have no clue.  ;)

sloede: In this command above, you do not set the `JULIA_DEPOT_PATH`, are you sure about that?

bgeihe: I'm not really sure about the implications. So I actually tried both, with and without.

sloede: I think I get it now, and you were right in the first place:

Since we *do* evaluate the elixirs in `Main`, the packages that are `include`d in the libelixirs actually do need to be loaded in the project that is currently active. I was wrong in my analysis back then.

Now, the question is, how to proceed here? I am not 100% sure that a *canonical* approach would look like, thus I asked on [Discourse](https://discourse.julialang.org/t/how-to-canonically-load-and-run-user-provided-functions-from-a-file/103180) in the hope that it will shed some light on this. In the meantime, I see three options for moving forward:
1. Add back the dependencies to the `Project.toml` (including the same compat bounds)
2. Load the libelixirs in the `LibTrixi` module
3. Remove `using Trixi, OrdinaryDiffEq` from the libelixirs.

I think options 2) and 3) are too restrictive, since they do not allow users to provide libelixirs with additional dependencies (maybe someone would like to use code from another third-party package in the future?). Thus I suggest to use 1) and maybe add a comment to the compat sections in `Project.toml` and `test/Project.toml`, stating that if compat bounds are adapted, it should be verified that they match the ones in the other file for packages that are present in both files.

bgeihe: Thanks for you ongoing efforts!

I would certainly be fine with 1!

But, out if interest, what is the downside of 2?
If a user would like to use another Package, she would need to add it to LibTrixi's Project.toml, and not our `libtrixi-julia` project, right?
(plus issues with overwriting you mentioned on Discourse)

sloede: > But, out if interest, what is the downside of 2?
If a user would like to use another Package, she would need to add it to LibTrixi's Project.toml, and not our `libtrixi-julia` project, right?

Correct. And there's the problem: We don't want people to have to modify LibTrixi.jl just to use a different package - that would be imho very non-Julian and in general very inconvenient.

But we probably want to add it to our documention that any package that one wants to use in an libelixir needs to be added *a priori* to the `libtrixi-julia` project.

bgeihe: I agree. Both are important points!

.github/workflows/ci.yml
```diff
@@ -173,6 +173,15 @@ jobs:
...
           LIBTRIXI_DEBUG: all
+      - name: Run julia tests with coverage
+        if: ${{ matrix.test_type == 'coverage' }}
+        uses: julia-actions/julia-runtest@v1
+        with:
+          coverage: true
+          project: ./LibTrixi.jl
+        env:
+          JULIA_DEPOT_PATH: ~/.julia
+          LIBTRIXI_DEBUG: all
```
sloede, line 184:
```suggestion
      - name: Run julia tests with coverage
        uses: julia-actions/julia-runtest@v1
        with:
          coverage: ${{ matrix.test_type == 'coverage' }}
          project: ./LibTrixi.jl
        env:
          JULIA_DEPOT_PATH: ~/.julia
          LIBTRIXI_DEBUG: all
```
I wonder if this works, and if yes, whether this might make more sense: Always run the tests, but only run with coverage reporting if it is enabled globally?

sloede (commented):
One last suggestion, then this LGTM!

.github/workflows/ci.yml
bgeihe: Nice idea! Will try.

.github/workflows/ci.yml
```diff
@@ -173,6 +173,14 @@ jobs:
               ../build/examples/trixi_controller_data_f . libelixir_p4est2d_dgsem_euler_sedov.jl
         env:
           LIBTRIXI_DEBUG: all
+      - name: Run julia tests with coverage
```
sloede, line 176:
```suggestion
      - name: Run Julia tests
```

sloede (dismissed):
LGTM! Great work!

.github/workflows/ci.yml
sloede: It seems to work 👍
/r

bgeihe: Seems to work. Great!

Just something I noticed: All packages are precompiled again in the Julia testing step. There is no way getting around this because the test happens in a completely different Julia project, right?

sloede: > There is no way getting around this because the test happens in a completely different Julia project, right?

Hm. No, not necessarily. I wonder whether it's got to do with the `--coverage` settings, i.e., if it needs to be recompiled if coverage is enabled? Maybe you can ask this on Discourse and CC me, would be interesting to hear what the community has to say...

.github/workflows/ci.yml
```diff
@@ -221,23 +229,27 @@ jobs:
               ../build/examples/trixi_controller_data_c . libelixir_p4est2d_dgsem_euler_sedov.jl
           JULIA_DEPOT_PATH=~/.julia \
               ../build/examples/trixi_controller_data_f . libelixir_p4est2d_dgsem_euler_sedov.jl
-      - name: Process coverage data
+      - name: Process C coverage data
         if: ${{ matrix.test_type == 'coverage' }}
         run: |
           cd libtrixi-julia
           lcov --directory ../build  --capture --output-file lcov.info
+      - name: Process julia coverage data
```
bgeihe, line 237:
```suggestion
      - name: Process Julia coverage data
        if: ${{ matrix.test_type == 'coverage' }}
```

sloede: Let's commit this once the current checks have passed and then merge immediately.

sloede (approved):

PR #79, opened 2023-08-16
Restructure source code

### LINKED ISSUE #77: Restructuring
Files / folders:
- [x] `trixi.h` is the public interface definition, it should not change
- [x] ~~Organize C-API implementation according to purpose (`setup.c`, `version.c`, `control.c`, `data.c`)~~
- [x] C-API (`api.c`), Fortran-API (`api.f90`), Julia-API (`api_c.jl`, `api_jl.jl`) remain in single files
- [x] ~~function pointers must be made available to multiple c-files. Move them to `function_pointers.h/c`?~~
- [x] General purpose routines like `print_and_die` are moved to `auxiliary.h/c`

Code:
- [x] all interface routines should start with `trixi_`
- [x] `trixi_version_*` is replaced by `trixi_version_library_*` to be consistent with `trixi_version_julia*`
- [x] `julia_eval_string` is renamed to `trixi_eval_julia` and used in examples

bgeihe:
Resolves #77


bgeihe:
Initial draft for restructured C API code.

bgeihe:
This is now ready, following the discussion in #77. See the updated description in #77.

src/auxiliary.h
```diff
@@ -0,0 +1,23 @@
...
+void print_and_die(const char* message, const char* func, const char* file, int lineno);
+
+// Function to determine debug level
+int show_debug_output();
+
+// Function to evaluate julia REPL string with exception handling
+jl_value_t* checked_eval_string(const char* code, const char* func, const char* file,
+                                int lineno);
+
+// Function to get and store function pointers from julia to c functions
```
sloede, line 20:
```suggestion
// Function to get and store function pointers from Julia to c functions
```
You also write "C" and not "c" for the language 😬

src/auxiliary.c
```diff
@@ -0,0 +1,122 @@
...
+        exit(1);
+    }
+
+    assert(result && "Missing return value but no exception occurred!");
+
+    return result;
+}
+
+
+// Function to get and store function pointers from julia to c functions
```
sloede, line 100:
```suggestion
// Function to get and store function pointers from Julia to C functions
```

src/auxiliary.c
```diff
@@ -0,0 +1,122 @@
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "auxiliary.h"
+
+
+// Default depot path *relative* to the project directory
+// OBS! If you change the value here, you should also update the default value of
+// `LIBTRIXI_JULIA_DEPOT` in `utils/libtrixi-init-julia` accordingly
```
sloede, line 10: Do we have a similar warning in `utils/libtrixi-init-julia`? If yes, does it need to be updated since the value has been moved to a different file?

sloede (changes requested):
This looks really, really great! Very good job on cleaning up the API and making sure everything is neat and tidy. I think this will make it much easier going forward to add new functions without having to think "where should this go", and easier for users of libtrixi 👍

Just a few minor suggestions, but then this is ready to go!

src/auxiliary.c
bgeihe: Sure! Thanks!

src/auxiliary.h
bgeihe: That's right. I try to find and change other occurrences referring to the language.

sloede (approved):
LGTM!

Issue #73, opened 2023-08-14
Improve C interface tests


bgeihe:
Right now our CI runs two examples (each in C and Fortran) and checks whether they finish without errors.

Do we want more?

- Unit tests? 
- Tests with random or perturbed data?
- Third party testing framework?

sloede:
IMHO, we probably want to keep our testing infrastructure as simple as possible. That is, it should be simple to add new tests (otherwise they will not be created) and simple to identify bugs (otherwise they will not be fixed).

At the moment, I am not 100% sure what our setup should/can look like. But from our experience with Trixi.jl, I'd say that we'll probably have more or less "integrated unit tests", e.g., something similar to our current controller examples but where we verify the result of each API call. I am not sure if we need a third-party testing framework for this, or if we are better off by rolling our own.

What I don't think we need is testing with random/perturbed data. Most of the functionality we want to test is regarding the interface, and there a correct answer is usually easy to detect. We do not need, e.g., to verify the results of simulations (since this is done already in Trixi.jl), but only things like transferring data to-and-fro, and that all API calls work as expected.

Does this help?

bgeihe:
Yes, thanks for your thoughts on this!

sloede:
There are some Fortran unit testing frameworks, as listed, e.g., [here](https://github.com/Beliavsky/Fortran-code-on-GitHub/blob/main/README.md#unit-testing) or [here](https://fortranwiki.org/fortran/show/Unit+testing+frameworks). However, I don't know any of them, so I can't give good recommendations.

We could also try to roll our own, e.g., to essentially code up 2-3 examples as tests, where we use the interface functions and verify the return values/functionality. Since we test the "external language <-> Julia" interaction already with the C code, we really only have to verify that the Fortran layer does not change anything while interacting with C, so I think it would be feasible to do it using manually coded tests.

PR #80, opened 2023-08-17
C API test using google test

bgeihe:
First part adressing #73


test/interface_c.cpp
```diff
@@ -0,0 +1,112 @@
...
+extern "C" int trixi_initialize_simulation(const char * libelixir);
+extern "C" void trixi_finalize_simulation(int handle);
+extern "C" void trixi_finalize();
+extern "C" int trixi_is_finished(int handle);
+extern "C" void trixi_step(int handle);
+extern "C" double trixi_calculate_dt(int handle);
+extern "C" int trixi_ndims(int handle);
+extern "C" int trixi_nelements(int handle);
+extern "C" int trixi_nvariables(int handle);
+extern "C" void trixi_load_cell_averages(double * data, int handle);
```
sloede, line 19: Why not
```c++
extern "C" {
  int trixi_version_major();
  ...
}
```
?

bgeihe: Because I started with
```
extern "C" int trixi_version_major();
```
and the resorted to a known solution. :blush:

.github/workflows/ci.yml
```diff
@@ -110,10 +113,9 @@ jobs:
           mkdir build
           cd build
           cmake .. -DCMAKE_INSTALL_PREFIX=../install \
-                   -DCMAKE_C_FLAGS="-cpp --coverage -O0" \
-                   -DCMAKE_Fortran_FLAGS="-cpp --coverage -O0" \
-                   -DCMAKE_EXE_LINKER_FLAGS="--coverage" \
-                   -DCMAKE_SHARED_LINKER_FLAGS="--coverage"
+                   -DCMAKE_BUILD_TYPE=Debug \
+                   -DTESTING=ON -DJULIA_PROJECT_PATH=../libtrixi-julia \
```
bgeihe, line 117: This activates building the tests. One has to specify the path to the Julia project such that Trixi.jl can run.

.github/workflows/ci.yml
```diff
@@ -110,10 +113,9 @@ jobs:
...
+                   -DCOVERAGE=ON
```
bgeihe, line 118: Option for turning on coverage reporting. This is nice for developers to locally get the report as well, but certainly not required for end users. I wonder if it should really be added.

The cmake module (see below) seem uncritical w.r.t. licensing.

test/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+
+    # enable warnings
+    target_compile_options( ${TARGET_NAME} PRIVATE -Wall -Wextra -Werror )
+
+    # pass julia project path
+    target_compile_definitions( ${TARGET_NAME} PRIVATE 
+                                JULIA_PROJECT_PATH=\"${JULIA_PROJECT_PATH}\" )
+
+    # TODO: add to installation?
+    #install( TARGETS ${TARGET_NAME} )
```
bgeihe, line 44: Should tests always run from build directory? Is there a reason to install them as well?

test/interface_c.cpp
```diff
@@ -0,0 +1,182 @@
...
+    // Initialize Trixi
+    trixi_initialize( julia_project_path, NULL );
+
+    // Execute correct Julia code
+    // NOTE: capturing stdout somehow does not work
+    trixi_eval_julia("println(\"Hello from Julia!\")");
+
+    // Execute erroneous Julia code
+    // NOTE: output before exit is somehow not captured here
+    EXPECT_DEATH(trixi_eval_julia("printline(\"Hello from Julia!\")"), "");
```
bgeihe, line 102: This was unexpected. 

Calling `trixi_eval_julia` with a wrong julia command results in error messages as expected, see
https://github.com/trixi-framework/libtrixi/blob/5dd0bbdef0428656c390c6b5ebb26d4e47c1f13f/src/auxiliary.c#L81-L92

However this output somehow cannot be captured using gtest's `testing::internal::CaptureStdout()/CaptureStderr()` (which is internal and not documented anyway)

Moreover, in the following, `jl_atexit_hook(1)` results in an error.

> atexit hook threw an error: ReadOnlyMemoryError()

which I do not understand. `exit(1)` is never reached.

In the end it does not really matter. The test fails as expected, but with unexpected last words.

.github/workflows/ci.yml
sloede: Sounds reasonable to me 🤷‍♂️

.github/workflows/ci.yml
sloede: If it is not required for CI testing, tbh, I'd leave it out, following [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). I don't see anyone running coverage tests locally _before_ pushing to the repo anyways. If you _do_ see a need for this, I just recommend to weigh it carefully against the potential overhead of having more code to maintain and more things that can possibly break 😅

test/CMakeLists.txt
sloede: Are we talking about the examples or about the tests? I think tests do not need to be installed, demos should be. But I have no hard opinion here.

test/interface_c.cpp
sloede: Phew. Good question. I don't know. If it works for all non-abnormal cases, let's just be happy about it and ignore this for now :-)

.github/workflows/ci.yml
bgeihe: Cool!

.github/workflows/ci.yml
bgeihe: Sounds reasonable.

I will just try to keep it in my head.

test/CMakeLists.txt
bgeihe: Tests!
I am very fine to leave them out.

test/interface_c.cpp
bgeihe: Great!

bgeihe:
Coverage is rising :smile: 

I do not know how to provoke this one:
https://github.com/trixi-framework/libtrixi/blob/bb9582c8d720e15a6940bb9edacbd00f0683290b/src/auxiliary.c#L116-L120
If a non-existing function pointer name is given, the execution of the julia command will already abort with an error message.

sloede:
> If a non-existing function pointer name is given, the execution of the julia command will already abort with an error message.

Yeah, this path you cannot easily check. Just leave it be for now.

.github/workflows/ci.yml
```diff
@@ -165,6 +173,30 @@ jobs:
               ../build/examples/trixi_controller_data_f . libelixir_p4est2d_dgsem_euler_sedov.jl
         env:
           LIBTRIXI_DEBUG: all
+      - name: Check error handling of examples
+        if: ${{ matrix.test_type == 'coverage' }}
+        continue-on-error: true
+        run: |
+          cd libtrixi-julia
+          ../build/examples/simple_trixi_controller_c
```
bgeihe, line 181: Concerning the examples:

We can do something like this to increase coverage of the examples.
But:
1. I could not quickly figure out how to continue execution after the first command has failed
2. It would be nice to check of the return codes are indeed 2 as expected
3. Isn't this overkill?

sloede: > 1. I could not quickly figure out how to continue execution after the first command has failed
> 2. It would be nice to check of the return codes are indeed 2 as expected

The bash shell in GitHub Actions [by default](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell) runs with `set -e`, i.e., the first non-zero return value of any command causes to exit prematurely and mark the step as a failure.

What you could try is to manually catch and compare the exit code, e.g., something like this
```suggestion
      - name: Check error handling of examples
        if: ${{ matrix.test_type == 'coverage' }}
        run: |
          cd libtrixi-julia
          set +e # disable early exit on non-zero exit code
          ../build/examples/simple_trixi_controller_c
          if [ $? -eq 2 ]; then
            echo "Correct exit code 👍"
            exit 0
          else
            echo "ERROR: expected exit code 2" >&2
            exit 1
          fi
```
Note that I have not yet tested the code, so please verify it works.

> 3. Isn't this overkill?

I don't know. Maybe it is, but OTOH, we want to keep coverage up up up 😬

bgeihe: Great, thanks for the hints! So it is save to assume bash and we do not need to be cross platform compatible?

sloede: In libtrixi, we explicitly only support Linux at the moment, so I think it's ok. `bash` is also the default for GHA workflows, if you look carefully in the link to the GH docs above. Otherwise, you can also explicitly specify `bash` as the shell, which works cross platform (though I doubt we will ever support Windows in the foreseeable future).

bgeihe: It worked!

CMakeLists.txt
```diff
@@ -34,11 +34,20 @@ list ( APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/" )
...
 find_package( Julia REQUIRED )
 
-
-
 # Find MPI
 find_package( MPI REQUIRED )
 
+# Find google test on demand
+option( TESTING "Build tests using google test" )
+if( TESTING )
```
sloede, line 42: Is `TESTING` a canonical keyword here? Otherwise, I'd suggest to use something more explicit such as `ENABLE_TESTS` or `ENABLE_TESTING` etc.

Also:
```suggestion
option( TESTING "Build tests using Google Test" )
if( TESTING )
```

CMakeLists.txt
```diff
@@ -34,11 +34,20 @@ list ( APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/" )
...
+    find_package( GTest REQUIRED )
+    if ( NOT DEFINED JULIA_PROJECT_PATH )
+        message( WARNING "JULIA_PROJECT_PATH not set, tests will not work.")
```
sloede, line 45: Why not make this an error right now? Is there any conceivable use case where you'd want to enable tests but be OK with them failing?

CMakeLists.txt
```diff
@@ -34,11 +34,20 @@ list ( APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/" )
...
+    else()
+        set( JULIA_PROJECT_PATH ${JULIA_PROJECT_PATH} CACHE PATH "Path to julia project.")
```
sloede, line 47:
```suggestion
        set( JULIA_PROJECT_PATH ${JULIA_PROJECT_PATH} CACHE PATH "Path to Julia project (typically `libtrixi-julia`).")
```
Maybe for clarity?

src/auxiliary.c
```diff
@@ -38,7 +38,10 @@ void update_depot_path(const char * project_directory, const char * depot_path)
 
         // Construct absolute path
         char absolute_path[PATH_MAX];
-        realpath(path, absolute_path);
+        char * ret = realpath(path, absolute_path);
```
sloede, line 41:
```suggestion
        const char * ret = realpath(path, absolute_path);
```
?

test/interface_c.cpp
```diff
@@ -0,0 +1,182 @@
...
+    void trixi_finalize_simulation(int handle);
+    void trixi_finalize();
+    int trixi_is_finished(int handle);
+    void trixi_step(int handle);
+    double trixi_calculate_dt(int handle);
+    int trixi_ndims(int handle);
+    int trixi_nelements(int handle);
+    int trixi_nvariables(int handle);
+    void trixi_load_cell_averages(double * data, int handle);
+    void trixi_eval_julia(const char * code);
```
sloede, line 22:
```suggestion
#include path/to/trixi.h
```
? (otherwise, you will always need to adapt this as well when adding/modifying functions)

test/interface_c.cpp
```diff
@@ -0,0 +1,182 @@
...
+    std::string version_string("");
+    version_string.append(std::to_string(major));
+    version_string.append(".");
+    version_string.append(std::to_string(minor));
+    version_string.append(".");
+    version_string.append(std::to_string(patch));
+    
+    EXPECT_STREQ(version_string.c_str(), trixi_version_library());
+
+    // Check julia packages version information
```
sloede, line 79:
```suggestion
    // Check Julia packages version information
```
here and elsewhere 😬

test/interface_c.cpp
```diff
@@ -0,0 +1,182 @@
...
+    std::string version_string_julia(trixi_version_julia());
+    EXPECT_NE(version_string_julia.find("OrdinaryDiffEq"), std::string::npos);
+
+    std::string version_string_julia_ext(trixi_version_julia_extended());
+    EXPECT_NE(version_string_julia_ext.find("StartUpDG"), std::string::npos);
+
+    // Finalize Trixi
```
sloede, line 86:
```suggestion
    // Finalize libtrixi
```
? (to make it clear that we are initializing the library, not a simulation?)

test/interface_c.cpp
```diff
@@ -0,0 +1,182 @@
...
+}
+
+
+TEST(CInterfaceTest, JuliaCode) {
+
+    // Initialize Trixi
+    trixi_initialize( julia_project_path, NULL );
+
+    // Execute correct Julia code
+    // NOTE: capturing stdout somehow does not work
```
sloede, line 97: If you don't want output, why don't you just execute something like `1+1 == 2` which does not generate output?

sloede (commented):
Just a few small comments, but otherwise this looks great!

sloede:
>95% coverage 🥳

CMakeLists.txt
bgeihe: There was no specific reason for TESTING. ENABLE_TESTING is better!

CMakeLists.txt
bgeihe: Actually not, I think.

test/interface_c.cpp
bgeihe: Ok!
This looks a little strange to me, but it works!

test/interface_c.cpp
sloede: OK, then let's just leave it as it is

PR #84, opened 2023-08-23
Update julia controller example

bgeihe:
I would like to have `simple_trixi_controller.jl` installed as well (and get the instructions in README.md right). It is however not really `bin/`.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -49,3 +49,6 @@ foreach ( EXAMPLE ${EXAMPLES} )
     install( TARGETS ${TARGET_NAME} )
 
 endforeach()
+
+# install the julia controller example as well
+install( PROGRAMS simple_trixi_controller.jl TYPE BIN )
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -51,4 +51,4 @@ foreach ( EXAMPLE ${EXAMPLES} )
 endforeach()
 
 # install the julia controller example as well
-install( PROGRAMS simple_trixi_controller.jl TYPE BIN )
+install( PROGRAMS simple_trixi_controller.jl DESTINATION share/libtrixi/examples/ )
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -51,4 +51,4 @@ foreach ( EXAMPLE ${EXAMPLES} )
 endforeach()
 
 # install the julia controller example as well
-install( PROGRAMS simple_trixi_controller.jl DESTINATION share/libtrixi/examples/ )
+install( FILES simple_trixi_controller.jl DESTINATION share/libtrixi/examples/ )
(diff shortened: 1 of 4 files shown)


examples/CMakeLists.txt
```diff
@@ -49,3 +49,6 @@ foreach ( EXAMPLE ${EXAMPLES} )
     install( TARGETS ${TARGET_NAME} )
 
 endforeach()
+
+# install the julia controller example as well
+install( PROGRAMS simple_trixi_controller.jl TYPE BIN )
```
sloede, line 54: Why move it to the `bin/` directory? Can you directly execute it? I understand your motivation, but it feels completely _off_ for me to put a file in `bin` that I cannot just execute with `./filename`, i.e., that has runnable content and is `chmod`ded with `+x`.

sloede: Why not leave it in `share/`?

> The /usr/share hierarchy is for all read-only architecture independent data files.

They likely did not consider these kind of non-executable script files when they wrote the [FHS specs](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html#ftn.idm236091847456), but it's the closest I found.

bgeihe: Ah, good reference. Thanks!

At the moment it is not installed at all. 
So where it exactly should it go?

- `share/libtrixi/`?
- `share/libtrixi/examples`?

libelexirs are installed to `share/libtrixi/LibTrixi.jl/examples`, but they are kind of a different thing.

sloede: `share/libtrixi/examples`?

bgeihe: Done!

examples/CMakeLists.txt
```diff
@@ -49,3 +49,6 @@ foreach ( EXAMPLE ${EXAMPLES} )
     install( TARGETS ${TARGET_NAME} )
 
 endforeach()
+
+# install the julia controller example as well
+install( PROGRAMS simple_trixi_controller.jl DESTINATION share/libtrixi/examples/ )
```
sloede, line 54:
```suggestion
install( FILES simple_trixi_controller.jl DESTINATION share/libtrixi/examples/ )
```
Otherwise, `+x` will be set on the file by [CMake](https://cmake.org/cmake/help/latest/command/install.html#programs):
> The `PROGRAMS` form is identical to the `FILES` form except that the default permissions for the installed file also include `OWNER_EXECUTE`, `GROUP_EXECUTE`, and `WORLD_EXECUTE`.

sloede (changes requested):

examples/CMakeLists.txt
bgeihe: Thanks for being so attentive!

sloede (approved):
LGTM!

PR #85, opened 2023-08-25
Running Google Test with MPI

bgeihe:
Some experimental code to run a gtest with MPI.

The good thing is, it seems to work within our CI.

The bad thing is that it fails.
What actually fails is however instructive. 

- The number of elements is 128 instead of 256. Given the fact that 2 processes were used, this seems plausible.
- Some computed cell average value is just 0. So it seems the solution data was not synchronized, which is plausible as I did not care about it.


sloede:
> * The number of elements is 128 instead of 256. Given the fact that 2 processes were used, this seems plausible.

Yes, this is good news 👍 However, it also shows that we should update this information from Trixi.jl in the `AnalysisCallback` to show the global number of elements, and not just the local number (which it does right now). It also shows that you need to make GTest MPI aware. Thus, you should possibly link it against MPI and initialize MPI within GTest, such that you also have access to `MPI_Comm_rank`, `MPI_Comm_size` etc. to figure out how to adjust your expectations (e.g., for the number of local elements).

Another option would be to set a special environment variable (e.g., `LIBTRIXI_NUM_RANKS`) and if it is set, get the information from there. This might be easier in the beginning, but eventually you will have to use actual MPI support, so maybe just go down this road immediately.
 
> * Some computed cell average value is just 0. So it seems the solution data was not synchronized, which is plausible as I did not care about it.

I'm not sure it's a necessarily synchronization problem: Each rank is supposed to compute the cell averages of their own rank. So when you're checking for the average in cell 929 (928 + 1) in https://github.com/trixi-framework/libtrixi/pull/85/files#diff-bf51675732cd01fd6ac5d3c6f3645b10163b540a95927b3c2b8dc15a529ee673R55, not only does the it give you the value of a different cell, but also of a different variable:

In the original case, you had 1600 entries per variable (Structure of Arrays layout), with 2 ranks you only have 800 on each rank. Thus, before you were checking the 929th density value, with 2 ranks you are checking the 129th x-velocity value.

Here's a good resource on using Google Test with MPI: https://elib.dlr.de/144193/1/Testing%20HPC%20C%2B%2B%20software%20with%20GoogleTest.pdf
Maybe this will give you some ideas how to add parallel testing with GTest?

It seems like also the t8code folks use it (see, e.g., https://github.com/DLR-AMR/t8code/blob/main/test/t8_gtest_main.cxx), so maybe this could be something to bring up at the next ADAPTEX meeting where t8code folks are in attendance?

bgeihe:
Thanks for the detailed analysis and the helpful hints!

bgeihe:
Just for reference:
https://github.com/DLR-SC/googletest_mpi
It seems a little outdated though

LibTrixi.jl/src/LibTrixi.jl
```diff
@@ -24,9 +25,12 @@ export trixi_step,
 export trixi_ndims,
        trixi_ndims_cfptr,
        trixi_ndims_jl
-export trixi_nelements,
-       trixi_nelements_cfptr,
-       trixi_nelements_jl
+export trixi_nelements_local,
+       trixi_nelements_local_cfptr,
+       trixi_nelements_local_jl
```
sloede, line 30: Without going into more details here, I'd recommend to not call it `nelements_local` but to stick to `nelements` and `nelements_global`. Most codes (that I know) are written such that they assume that all values are local unless noted otherwise, since otherwise you'd have to add the `_local` suffix in too many places.

sloede:
I don't get the errors: For example, https://github.com/trixi-framework/libtrixi/actions/runs/6038153366/job/16383897333?pr=85#step:19:382 compares to values, `0.0028566952356658794` and `0.0028566952356658751`, which have a difference of `-4.336808689942018e-18`. Maybe your tolerances are too tight?

bgeihe:
> Maybe your tolerances are too tight?

Exactly! I had used `EXPECT_DOUBLE_EQ`, which according to the docs
> Verifies that the two double values val1 and val2 are approximately equal, to within 4 ULPs from each other.
Units in the Last Place (ULPs), I had not heard of before.

I now use `EXPECT_NEAR` with an absolute tolerance of 1e-14 (values to compare are in the range [0.1, 0.9])

bgeihe:
Julia test for the new `trixi_nelemens_global` was missing.
Coverage should recover.

test/c/simulation.cpp
```diff
@@ -0,0 +1,131 @@
...
+    MPI_Comm comm = MPI_COMM_WORLD;
+
+    int rank;
+    MPI_Comm_rank(comm, &rank);
+
+    int nranks;
+    MPI_Comm_size(comm, &nranks);
+
+    // Initialize libtrixi
+    trixi_initialize( julia_project_path, NULL );
```
sloede, line 33:
```suggestion
    trixi_initialize(julia_project_path, NULL);
```
I think you've been somewhat inconsistent about whether to include spaces inside parens, so maybe you need to make up your mind and just use one style everywhere 😬

test/c/simulation.cpp
```diff
@@ -0,0 +1,131 @@
...
+            EXPECT_NEAR(cell_averages[33+2*nelements],  0.14037267400591, 1e-14);
+            EXPECT_NEAR(cell_averages[34+  nelements],  0.14037267400591, 1e-14);
+        }
+    }
+    else {
+        FAIL() << "Test cannot be run with " << nranks << " ranks.";
+    }
+    
+    // Finalize Trixi simulation
+    trixi_finalize_simulation( handle );
```
sloede, line 120:
```suggestion
    trixi_finalize_simulation(handle);
```

sloede (commented):
Minor suggestions, but other than that, this looks great!

sloede (approved):

PR #89, opened 2023-08-28
Add Fortran interface tests

### LINKED ISSUE #82: Fortran Tests
Unfortunately Fortran is not natively supported by google test. A way around would be to write a Fortran program testing everything that is currently covered by the C-tests, write an interface to this function, and call it from a C++ google test.
It feels a little strange though, because we would call an interface function to Fortran, which would then call interface functions to C.

Alternatively, provided we keep running our examples, we could make sure that the examples use all of the provided Fortran functions.

The latter however seems not straightforward as apparently the Fortran interface calls are not considered in the coverage reports.

_Originally posted by @bgeihe in https://github.com/trixi-framework/libtrixi/issues/73#issuecomment-1686532008_

bgeihe:
Resolves #82


bgeihe:
Finally!

Coverage is looking good as well. New "issues":

1.
  ```
  void trixi_initialize(const char * project_directory, const char * depot_path) {
      // Prevent double initialization and initialization after finalization
      if (is_initialized) {
          print_and_die("trixi_initialize invoked multiple times", LOC);
      }
      if (is_finalized) {
          print_and_die("trixi_initialize invoked after trixi_finalize", LOC);
      }
  ```
  The second safeguard is never reached, because `is_finalized` can only be set in `finalize`, and `finalize` can only run if  `is_initialized` is set.

2. As expected, the Fortran interface definition is not captured.

3. Multi-Line statements in Fortran (with &) are not correctly handled.

4. `test/fortran` is checked for coverage (which is not really useful),
`test/c` is not (maybe due to gtest magic creating executables out of nowhere?)

sloede:
> The second safeguard is never reached

Good point! Can you please remove the second check?

> As expected, the Fortran interface definition is not captured. 

Too bad, but nothing we can do here, I guess, but to remain vigilant ourselves. 

> test/fortran is checked for coverage (which is not really useful),
test/c is not (maybe due to gtest magic creating executables out of nowhere?) 

Can you exclude test from coverage reporting?

bgeihe:
All done! Excluding feels a little like cheating, but coverage increased!

sloede (approved):
Great work 🥳

PR #91, opened 2023-08-28
Specify C11 and Fortran 2018 language standards

### LINKED ISSUE #90: Set C and Fortran standard versions
We should explicitly set the C/Fortran versions we want to compile against such that there are no surprises about features being supported differently on different compilers, or ambiguity about which features we use or need.

Since requiring an arbitrarily recent language standard limits the choice of compilers, I suggest to try to play it safe for now and require C11 and Fortran ~~2008~~ 2018 (if those actually work).

**EDIT**: We actually need Fortran 2018 😱

sloede:
Fixes #90.


bgeihe (approved):

sloede:
For reference, Julia uses `-std=gnu11` [as well](https://github.com/JuliaLang/julia/blob/8ace9c5a833ef0075ffc9238ea65ed6f18dcfef8/Make.inc#L491), thus it makes sense that using a stricter standard in files that include `julia.h` failed when tested in this PR.

PR #93, opened 2023-08-29
Miscellaneous CI improvements

sloede:
This PR combines various small CI improvements that hopefully make it easier to understand or adapt the main CI testing config file. Since a lot in the file has changed, I recommend for a review to either directly read the entire new file and not look at the diff, or to just step through all commit diffs manually to understand the changes that have been performed.

Here's a summary of the main modifications:
* Throw out unused configurations (Windows, macOS, Valgrind)
* Enable caching of t8code installation (to save ~3 minutes)
* Simplify the job configuration by using a simpler matrix
* Execute Google Tests also in 'regular' runs

Some of the changes might be too far reaching, so feel free to comment/question these choices.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Show Julia version information
         run: julia -e 'using InteractiveUtils; versioninfo(verbose=true)'
 
-      - name: Install dependencies available as Ubuntu packages
+      - name: Install dependencies available with `apt-get`
         # - MPI
         # - HDF5
         # - Google Test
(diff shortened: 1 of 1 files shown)


bgeihe (approved):
This really looks nice!

Just one thing: What about the julia depot path? Our julia project directory is `libtrixi_julia`, so by default the depot would be at `libtrixi-julia/julia-depot`.

sloede:
> Just one thing: What about the julia depot path? Our julia project directory is `libtrixi_julia`, so by default the depot would be at `libtrixi-julia/julia-depot`.

You're right. We have this super special logic that reacts on the path being set. Hopefully fixed in 8c2917c.

bgeihe (approved):

bgeihe:
Thanks again!

> We have this super special logic that reacts on the path being set.

...which we could still change if it turns out to be rather confusing then helpful.

PR #95, opened 2023-08-30
Speed up CI testing by precompiling only once

sloede:


sloede:
It's a small improvement only, but it seems like precompilation is indeed only happening once anymore.

bgeihe (approved):
A truly heroic endeavor!

Too sad the gain does not feel larger (I have not checked). 
I would have though so, because the runtime increased so much once I added the Julia tests.

PR #96, opened 2023-08-31
Split C/Fortran tests and colorize Google Test output in CI tests

sloede:


bgeihe (approved):
Nice! I did not know about this env.

PR #97, opened 2023-08-31
Make t8code available in libtrixi

bgeihe:


cmake/FindT8code.cmake
sloede, line 1: Where does this file come from? Did you create it yourself or copy it from t8code?

bgeihe: Own work, which might very well fail.

LibTrixi.jl/src/api_jl.jl
```diff
@@ -129,3 +129,9 @@ function trixi_load_cell_averages_jl(data, simstate)
 
     return nothing
 end
+
+
+function trixi_get_t8code_mesh_jl(simstate)
+    mesh, _, _, _ = Trixi.mesh_equations_solver_cache(simstate.semi)
+    return mesh.forest
```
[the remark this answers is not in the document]
sloede: Maybe just call it `trixi_get_t8code_forest_jl` for now

cmake/FindT8code.cmake
sloede: Why not create a PR to the t8code repo and add this cmake file to t8code itself (while keeping this one here for now)? If they accept it, you have an official version, and if they find issues with it, it will help you make your script better here too

LibTrixi.jl/src/api_jl.jl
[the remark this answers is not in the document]
bgeihe: Yes, I changed that!

cmake/FindT8code.cmake
bgeihe: Good idea!

bgeihe:
At the moment we get
```
[libsc 0] Abort: Assertion '((sc_mempool_t *) ts_context)->elem_count == 0' 
[libsc 0] Abort: src/t8_schemes/t8_default/t8_default_common/t8_default_common_cxx.cxx:54 
[libsc 0] Abort
```
when finalizing Trixi. This need further investigation.

bgeihe:
@sloede If time permits, I would like to hear your thoughts on this (see also the description).

sloede:
Some initial thoughts from my side:

>     * At the moment there is not much functionality. Actually only
>       `t8_forest_t trixi_get_t8code_mesh(int handle);`
>       However this could be enough. A user would get the mesh from libtrixi, and then use the C/Fortran interface of t8code. We would use the T8code.jl interface internally.

Currently it is hard to say which functionality will be needed in practice and how it is going to be used. For example, I cannot imagine that it will be OK for Trixi.jl to "own" and manage the mesh, but other actors to also modify it, since that will break many assumptions in Trixi.jl. Thus maybe it is ok for now to have such a function, although given the t8code docs in the wiki, maybe for completeness one should provide
```c
t8_forest_t trixi_get_t8code_forest(int handle);
t8_cmesh_t trixi_get_t8code_cmesh(int handle);
```
I would definitely mark those API functions as experimental for now, such that we can remove them without having to create a breaking release.

>     * At the moment we need to include `t8.h` only because of the type `t8_forest_t`. Linking the library is only required for the example. On the other hand, we require our users to install t8code anyway.

I've seen this, and I agree, this is somewhat annoying. Would it be an option if we just give the definitions of these opaque pointers ourselves? That is, just put
```c
typedef struct t8_cmesh *t8_cmesh_t;
typedef struct t8_forest *t8_forest_t;
```
in our own files? I guess it would be very bad style though 😬 Another option to avoid the dependency on t8code would be to return `void*` instead, which is also not the cleanest of clean solutions.

Since we are not sure how we are going to go with this, I recommend to just add the two typedefs to our own headers to avoid forcing users to have t8code installed before compiling libtrixi. We should mark this with comments in the source code and possibly create an issue for this to keep track of the question, but that seems to me the "easiest" approach for now. Once it is clear what kind of interface is required, we can find a more elegant, permanent solution.

>     * There is no Fortran example, yet, as this would require the t8code Fortran interface, which is still under development.

That's OK I guess. Just leave a comment saying as much.

Hope that helps!

sloede:
Regarding declaring the typedefs `t8_cmesh_t` and `t8_forest_t` yourself:

Since I was curious, I just investigated this question a little further (from a C language legality POV), and found https://stackoverflow.com/a/8367810/1329844. This says that for C11, redefining a typedef is legal, as long as it is defined to the same thing. The C11 standard (to be precise: [the final draft](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf)), Section 6.7 3 says:
> a typedef name may be redefined to denote the same type as it currently does, provided that type is not a variably modified type

The constraint "variably modified type" does not apply, since it essentially means "variable length array or types derived thereof" (e.g., a pointer to a VLA), and VLAs can not be part of a struct.

To be on the even safer side, you could consider wrapping the type declaration in the header guards used by the good people of t8code, e.g., as
```c
#if !defined(T8_H) && !defined(T8_FOREST_GENERAL_H)
t8_forest_t trixi_get_t8code_forest(int handle);
#endif
#if !defined(T8_H) && !defined(T8_CMESH_H)
t8_cmesh_t trixi_get_t8code_cmesh(int handle);
#endif
```
If you put this in `trixi.h`, it should make reasonably sure that it can be used with pre-C11 compilers, as it will safeguard against prior inclusion of the main t8code header or the specific header file with the typedefs.

examples/CMakeLists.txt
```diff
@@ -27,14 +31,21 @@ foreach ( EXAMPLE ${EXAMPLES} )
     # set libraries to link
     target_link_libraries(
         ${TARGET_NAME}
-        PRIVATE MPI::MPI_${EXAMPLE_LANG} ${PROJECT_NAME} ${PROJECT_NAME}_tls
+        PRIVATE MPI::MPI_${EXAMPLE_LANG}
+                ${PROJECT_NAME} ${PROJECT_NAME}_tls
```
bgeihe, line 35:
```suggestion
        PRIVATE MPI::MPI_${EXAMPLE_LANG} ${PROJECT_NAME} ${PROJECT_NAME}_tls
```

test/c/interface_c.cpp
```diff
@@ -1,4 +1,5 @@
 #include <gtest/gtest.h>
+#include <mpi.h>
```
bgeihe, line 2:
```suggestion
```

src/trixi.h
```diff
@@ -32,6 +32,16 @@ int trixi_nvariables(int handle);
 double trixi_calculate_dt(int handle);
 void trixi_load_cell_averages(double * data, int handle);
 
+// T8code
+#if !defined(T8_CMESH_H)
+typedef struct t8_cmesh *t8_cmesh_t;
+#endif
+#if !defined(T8_FOREST_GENERAL_H)
+typedef struct t8_forest *t8_forest_t;
+#endif
```
sloede, line 41:
```suggestion
#if !defined(T8_H) && !defined(T8_CMESH_H)
typedef struct t8_cmesh *t8_cmesh_t;
#endif
#if !defined(T8_H) && !defined(T8_FOREST_GENERAL_H)
typedef struct t8_forest *t8_forest_t;
#endif
```
I'd also include the `T8_H` header guard just to be on the safe side. They might rename the file where these typedefs live, but they most likely will not rename `t8.h`.

sloede (changes requested):

src/trixi.h
bgeihe: Ok! I now understand why you initially suggested this!

LibTrixi.jl/src/api_c.jl
```diff
@@ -354,3 +354,45 @@ end
...
+Return t8code cmesh of the current T8codeMesh.
+
+!!! warning "Experimental"
+    The interface to t8code is experimental and implementation details may change at any
+    time without warning.
+"""
+function trixi_get_t8code_cmesh end
+
+Base.@ccallable function
+    trixi_get_t8code_cmesh(simstate_handle::Cint)::Ptr{Trixi.t8_cmesh}
```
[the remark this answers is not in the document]
bgeihe: Ok! I now understand why you suggested this initially.

sloede (approved):
LGTM! And thanks for bearing with me!

PR #100, opened 2023-09-03
Fix version test

bgeihe:
Compare version only up to `-<pre-release>`


sloede:
Tests are still failing, did you really fix everything?

bgeihe:
No, please ignore the request.

sloede (approved):
As discussed via Slack, it would be great if you can make the Fortran tests also work for version numbers with double-digit (or more) version components.

PR #102, opened 2023-09-04
Fortran version test without fixed integer width

bgeihe:


sloede (approved):
LGTM! That means we support up to `XXX.YYY.ZZZ`?

bgeihe:
Exactly!
Plus `-<prerelease>` which is however not further checked.

PR #107, opened 2023-09-06
Add DOI and referencing information to the docs

### LINKED ISSUE #104: Create Zenodo DOI
Now that we have releases, we should add a Zenodo DOI.

sloede:
Resolves #104.


bgeihe (approved):
Really nice to have an own DOI!

PR #110, opened 2023-09-11
Make TagBot work again (at all)

sloede:
With https://github.com/JuliaRegistries/TagBot/pull/279 merged, we need to wait for a new version of the TagBot action to be tagged (no pun intended). Once this has happened, we should be able to use TagBot for automatic new release creation.


sloede:
A new version of TagBot (v1.16.0) has been released - I think we should be good to go!

sloede:
@bgeihe Once this has been merged, you could also create a new release v0.1.2 that incorporates the fixes to HDF5.

bgeihe (approved):

PR #111, opened 2023-09-12
Use set_preferences to set system hdf5

bgeihe:
HDF5 0.17.0 has arrived and we have to switch to using `set_preferences`.

### DIFF
--- a/utils/libtrixi-init-julia
+++ b/utils/libtrixi-init-julia
@@ -320,7 +320,7 @@ if [ "$CI" = "true" ] && [ -n "$LD_PRELOAD" ]; then
   store_ld_preload="$LD_PRELOAD"
   LD_PRELOAD=""
 fi
-JULIA_HDF5_PATH="$hdf5_libdir" JULIA_DEPOT_PATH="$julia_depot" $julia_exec --project=. \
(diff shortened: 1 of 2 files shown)


utils/libtrixi-init-julia
```diff
@@ -314,17 +314,24 @@ echo
 
 # Set system HDF5 library
 echo "Use system HDF5 for Julia... "
+# Disable preloading libcurl.so if it has been set when running within the GitHub runner,
+# since it does not work with the JLL-provided libhdf5.so
+if [ "$CI" = "true" ] && [ -n "$LD_PRELOAD" ]; then
+  store_ld_preload="$LD_PRELOAD"
+  LD_PRELOAD=""
+fi
 JULIA_HDF5_PATH="$hdf5_libdir" JULIA_DEPOT_PATH="$julia_depot" $julia_exec --project=. \
```
sloede, line 323:
```suggestion
JULIA_DEPOT_PATH="$julia_depot" $julia_exec --project=. \
```
With the preferences setup working, setting this variable should not be necessary anymore

sloede (changes requested):

sloede (approved):

PR #115, opened 2023-09-18
Store version info at compile time

sloede:
It was an oversight of myself to fill the version info arrays at *runtime* instead of *compile time*. Since the module gets recompiled whenever one of its (transitive) dependencies changes, this should always be up-to-date anyways.

This change hopefully fixes some issues encountered when trying to build libtrixi with PackageCompiler.jl.


sloede:
The remaining [uncovered line](https://coveralls.io/builds/62722727/source?filename=LibTrixi.jl%2Fsrc%2FLibTrixi.jl#L84) is hard to test, since it does not seem to occur during regular testing. I'll just go ahead and merge this anyways.

PR #116, opened 2023-09-18
Update documenter to v1

sloede:


sloede:
Looking good to me with only the issue of the relaxed link checking, for which an issue will be created.

PR #126, opened 2023-09-27
[WIP] Add getter function for `P4estMesh`

sloede:
Add capabilities for `P4estMesh` simulations imilar to `trixi_get_t8code_forest`.


sloede:
Note that this PR is still incomplete, since there is no easy way to fake the pointers to the `p4est_t` data structure. Unlike in t8code, `p4est_t` is actually a struct and not an opaque pointer. Thus we'd need to include the p4est headers, making p4est a hard requirement for even compiling the library. I'm not saying we should never do that, but I haven't found the time yet.

PR #128, opened 2023-09-29
Make PackageCompiler.jl compilation selectable in cmake

### LINKED ISSUE #124: Make PackageCompiler library selectable via cmake
> No, I’d say that it’s either PackageCompiler or our C library. I would treat them as mutually exclusive, not a composition. So you either use our C library or the PC approach during installation. For that, we also need a new CMakeLists.txt (or better yet, a selector whether you want to use the C shim or PC) and then fix the installation part accordingly. What’s also missing right now is the ability to use the Fortran examples.

_Originally posted by @sloede in https://github.com/trixi-framework/libtrixi/pull/76#discussion_r1335655520_

**EDIT** (by @sloede): In this course, one should also introduce the ability to do full out-of-source builds. That is, no compilation artifacts etc. should end up in `LibTrixi.jl/lib`.

bgeihe:
Resolves #124 

- Introduces a new option `USE_PACKAGE_COMPILER`
- Requires `JULIA_PROJECT_PATH` being set (as does testing)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,9 @@ jobs:
       - name: Show Julia version information
         run: julia -e 'using InteractiveUtils; versioninfo(verbose=true)'
 
+      - name: Update `apt-get`
+        run: sudo apt-get update
+
       - name: Install dependencies available with `apt-get`
         # - MPI
         # - HDF5
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -186,6 +186,17 @@ jobs:
           cd build
           make -j2
 
+      - name: Test external CMake project
+        if: ${{ matrix.test_type == 'regular' }}
+        run: |
+          cd examples/external
+          ./build.sh
+          mpirun -n 2 ./build/simple_trixi_controller_c \
+              ../../libtrixi-julia \
+              ../../LibTrixi.jl/examples/libelixir_p4est2d_dgsem_euler_sedov.jl
+        env:
+          LIBTRIXI_DEBUG: all
+
       - name: Prepare coverage reporting
         if: ${{ matrix.test_type == 'coverage' }}
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,10 +220,10 @@ jobs:
       - name: Run examples
         if: ${{ matrix.test_type == 'package-compiler' }}
         run: |
-          cd examples
+          cd build/examples
           mpirun -n 2 simple_trixi_controller_c \
-              ../libtrixi-julia \
-              ../LibTrixi.jl/examples/libelixir_p4est2d_dgsem_euler_sedov.jl
+              ../../libtrixi-julia \
+              ../../LibTrixi.jl/examples/libelixir_p4est2d_dgsem_euler_sedov.jl
         env:
           LIBTRIXI_DEBUG: all
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,17 +170,21 @@ jobs:
               --julia-depot ~/.julia \
               --force
 
-      - name: Build libtrixi using PackageCompiler
+      - name: Configure (test_type == 'package-compiler')
         if: ${{ matrix.test_type == 'package-compiler' }}
         run: |
-          cd LibTrixi.jl/lib
-          make
+          mkdir build
+          cd build
+          cmake .. -DCMAKE_INSTALL_PREFIX=../install \
+                   -DCMAKE_BUILD_TYPE=Debug \
+                   -DUSE_PACKAGE_COMPILER=ON \
+                   -DLIBTRIXI_JULIA_PROJECT=../libtrixi-julia
 
-      - name: Build example
+      - name: Build
         if: ${{ matrix.test_type == 'package-compiler' }}
         run: |
-          cd examples
-          make -f MakefileCompiled LIBTRIXI_PREFIX=$PWD/../LibTrixi.jl/lib/build
+          cd build
+          make -j2
 
       - name: Prepare coverage reporting
         if: ${{ matrix.test_type == 'coverage' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,7 +178,7 @@ jobs:
           cmake .. -DCMAKE_INSTALL_PREFIX=../install \
                    -DCMAKE_BUILD_TYPE=Debug \
                    -DUSE_PACKAGE_COMPILER=ON \
-                   -DLIBTRIXI_JULIA_PROJECT=../libtrixi-julia
+                   -DLIBTRIXI_JULIA_PROJECT=$PWD/../libtrixi-julia
 
       - name: Build
         if: ${{ matrix.test_type == 'package-compiler' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
           cmake .. -DCMAKE_INSTALL_PREFIX=../install \
                    -DCMAKE_BUILD_TYPE=Release \
                    -DT8CODE_PREFIX=$PWD/../t8code-local/prefix \
-                   -DENABLE_TESTING=ON -DLIBTRIXI_JULIA_PROJECT=../libtrixi-julia
+                   -DENABLE_TESTING=ON -DJULIA_PROJECT_PATH=../libtrixi-julia
 
       - name: Configure (test_type == 'coverage')
         if: ${{ matrix.test_type == 'coverage' }}
@@ -113,7 +113,7 @@ jobs:
                    -DCMAKE_Fortran_FLAGS="-cpp --coverage -O0" \
                    -DCMAKE_EXE_LINKER_FLAGS="--coverage" \
                    -DCMAKE_SHARED_LINKER_FLAGS="--coverage" \
-                   -DENABLE_TESTING=ON -DLIBTRIXI_JULIA_PROJECT=../libtrixi-julia
+                   -DENABLE_TESTING=ON -DJULIA_PROJECT_PATH=../libtrixi-julia
 
       - name: Build
         if: ${{ matrix.test_type == 'regular' || matrix.test_type == 'coverage' }}
@@ -178,7 +178,7 @@ jobs:
           cmake .. -DCMAKE_INSTALL_PREFIX=../install \
                    -DCMAKE_BUILD_TYPE=Debug \
                    -DUSE_PACKAGE_COMPILER=ON \
-                   -DLIBTRIXI_JULIA_PROJECT=$PWD/../libtrixi-julia
+                   -DJULIA_PROJECT_PATH=$PWD/../libtrixi-julia
 
       - name: Build
         if: ${{ matrix.test_type == 'package-compiler' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
           cmake .. -DCMAKE_INSTALL_PREFIX=../install \
                    -DCMAKE_BUILD_TYPE=Release \
                    -DT8CODE_PREFIX=$PWD/../t8code-local/prefix \
-                   -DENABLE_TESTING=ON -DJULIA_PROJECT_PATH=../libtrixi-julia
+                   -DENABLE_TESTING=ON -DLIBTRIXI_JULIA_PROJECT=../libtrixi-julia
 
       - name: Configure (test_type == 'coverage')
         if: ${{ matrix.test_type == 'coverage' }}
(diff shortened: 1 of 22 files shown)


bgeihe:
So far compilation in an out-of-source build directory seems to work.

Installation and examples are still missing.

LibTrixi.jl/lib/build.jl
```diff
@@ -54,7 +54,7 @@ header_files = [joinpath(dirname(dirname(@__DIR__)), "src", "trixi.h")]
 # Name of the files that include the initialization functions:
 # - `init.c` contains `trixi_initialize`/`trixi_finalize` for API compatibility
 # - the other file contains the `init_julia`/`shutdown_julia` functions from PackageCompiler
-julia_init_c_file = ["init.c", PackageCompiler.default_julia_init()]
+julia_init_c_file = [joinpath(@__DIR__, "init.c"), PackageCompiler.default_julia_init()]
```
sloede, line 57: Note that PackageCompiler.jl will create the object file `init.o` right next to the input file `init.c`. Thus I think we need to copy the `init.c` first to our build tree somewhere, otherwise we end up polluting the source tree.

CMakeLists.txt
```diff
@@ -60,74 +60,92 @@ if( ENABLE_TESTING )
...
+    endif()
 
+    # no sources
+    add_library( ${PROJECT_NAME} INTERFACE build-pc/lib/libtrixi.so )
+    add_custom_command( OUTPUT build-pc/lib/libtrixi.so
+                        COMMENT "Building ${PROJECT_NAME} with PackageCompiler.jl..."
+                        COMMAND ${JULIA_EXECUTABLE}
+                                --project=${CMAKE_SOURCE_DIR}/LibTrixi.jl/lib
+                                ${CMAKE_SOURCE_DIR}/LibTrixi.jl/lib/build.jl
+                                ${JULIA_PROJECT_PATH}
```
sloede, line 77: I think `JULIA_PROJECT_PATH` is not an optimal name. It *hints* at being something used by Julia, but it is not (we use it in libtrixi). What about renaming it (here and elsewhere) to `LIBTRIXI_JULIA_PROJECT`? This is similar enough to Julia's own [`JULIA_PROJECT`](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_PROJECT) such that those familiar with Julia will know what this is about, while clearly indicating that this is not used by Julia directly but just by us?

CMakeLists.txt
```diff
@@ -60,74 +60,92 @@ if( ENABLE_TESTING )
...
+# Optionally use PackageCompiler.jl to build standalone libtrixi.so
+option( USE_PACKAGE_COMPILER "Build standalone libtrixi.so using PackageCompiler.jl" )
+if( USE_PACKAGE_COMPILER )
+    if ( NOT DEFINED JULIA_PROJECT_PATH )
+        message( FATAL_ERROR "JULIA_PROJECT_PATH needs to be set for PackageCompiler.jl.")
+    endif()
 
+    # no sources
+    add_library( ${PROJECT_NAME} INTERFACE build-pc/lib/libtrixi.so )
+    add_custom_command( OUTPUT build-pc/lib/libtrixi.so
```
sloede, line 72: Should we add other files to the output as well, e.g.,
```
build-pc/include/trixi.h
build-pc/share/julia/LIBTRIXI_VERSION
build-pc/share/julia/LocalPreferences.toml
build-pc/share/julia/Project.toml
```
?

I think it is probably too hard/not useful enough to also include the other outputs (such as all the libraries that are bundled), since they are just copy-pasted into the bundle.

bgeihe: Actually I do not know (yet) as I do not really understand what is going on in cmake here (yet).

My intention was to define a target similar to what we have when building `libtrixi.so` from sources via the usual C toolchain. 

Installing could work by just copying the share, lib, include directories explicitly. 

Linking the examples so far works by 
```
target_link_libraries(
        ${TARGET_NAME}
        ${PROJECT_NAME}
    )
```
and ideally this (i.e. just specifying `${PROJECT_NAME}` for the library) could be reused.

sloede: OK, makes sense. Let's keep it as it is for now and (Maybe) reconsider this in the future. I think adding the other missing functionality (building examples etc) is much more important.

CMakeLists.txt
```diff
@@ -80,15 +80,18 @@ if( USE_PACKAGE_COMPILER )
     # Define PackageCompiler.jl output file
     set( PC_LIBTRIXI_SO ${CMAKE_BINARY_DIR}/prefix-pc/lib/libtrixi.so )
 
-    # Add a library target (libtrixi), but only for Fortran module
+    # Add a library target (libtrixi)
+    # - for Fortran module (therefore OBJECT library)
+    # - for generating a rule to build libtrixi.so
     add_library( ${PROJECT_NAME} OBJECT
         src/api.f90
+        ${PC_LIBTRIXI_SO}
```
sloede, line 88: Why is this necessary?

bgeihe: Good question!

Short answer: it does not work otherwise

Long answer:

> This [add_custom_command] defines a command to generate specified OUTPUT file(s). A target created in the same directory (CMakeLists.txt file) that specifies any output of the custom command as a source file is given a rule to generate the file using the command at build time.

So, if I understand this correctly, the output (libtrixi.so) has to appear as a source file in order to create to a build rule for it.

sloede: You might need to add an `add_custom_target` (https://dev.to/iblancasa/learning-cmake-3-understanding-addcustomcommand-and-addcustomtarget-43gp) or otherwise use a full path in `target_link_libraries` (https://stackoverflow.com/a/44084377/1329844). Not sure if these apply here, but maybe it helps.

Maybe instead of adding it as a source, you could also add the `${PC_LIBTRIXI_SO}` through `target_link_libraries` (https://cmake.org/cmake/help/latest/command/target_link_libraries.html#command:target_link_libraries)?

I am not sure what the correct setup is for this, maybe we have to ask on the CMake Discourse forum. The challenge here is imho that usually CMake expects to build the library itself, or to use an existing library somewhere, but not this mixture that we have 🤷‍♂️

bgeihe: Thanks for the pointers! I will try `add_custom_target`.

Just for explanation: `${PC_LIBTRIXI_SO}` actually is a full path, and `target_link_libraries` is already used (and apparently required).

.github/workflows/ci.yml
```diff
@@ -173,17 +173,21 @@ jobs:
...
-          make
+          mkdir build
+          cd build
+          cmake .. -DCMAKE_INSTALL_PREFIX=../install \
+                   -DCMAKE_BUILD_TYPE=Debug \
+                   -DUSE_PACKAGE_COMPILER=ON \
+                   -DJULIA_PROJECT_PATH=$PWD/../libtrixi-julia
 
-      - name: Build example
+      - name: Build
```
sloede, line 186:
```suggestion
      - name: Build (test_type == 'package-compiler')
```
For consistency?

CMakeLists.txt
```diff
@@ -56,78 +56,131 @@ if( ENABLE_TESTING )
...
 
-# Version info
-set_target_properties ( ${PROJECT_NAME} PROPERTIES VERSION ${PROJECT_VERSION} )
+    # Include directories, private
+    target_include_directories ( ${PROJECT_NAME} PRIVATE src )
 
-# Version info for the shared object
-set_target_properties ( ${PROJECT_NAME} PROPERTIES SOVERSION ${PROJECT_VERSION_MAJOR} )
+    # Version info
+    set_target_properties ( ${PROJECT_NAME} PROPERTIES VERSION ${PROJECT_VERSION} )
```
sloede, line 129: Why `VERSION` and not `SOVERSION` anymore?

CMakeLists.txt
```diff
@@ -56,78 +56,131 @@ if( ENABLE_TESTING )
...
+        message( NOTICE "Testing is not supported when PackageCompiler is used.")
+    endif()
 
-# Include directories, private
-target_include_directories ( ${PROJECT_NAME} PRIVATE src )
+    # Copy source with initialization routines; an object file will be created
+    file( MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/build-pc )
+    file( COPY_FILE ${CMAKE_SOURCE_DIR}/LibTrixi.jl/lib/init.c
+                    ${CMAKE_BINARY_DIR}/build-pc/init.c
+                    ONLY_IF_DIFFERENT )
```
sloede, line 81: `COPY_FILE` requires CMake v3.21, while we currently only require v3.9 (IIRC). Maybe use [`configure_file`](https://cmake.org/cmake/help/latest/command/configure_file.html#command:configure_file) instead?

sloede (changes requested):
LGTM overall, thanks a lot for doing this! It would be great if you could also document how to enable the PC build in the README and/or docs. I've left a few minor notes as well.

CMakeLists.txt
bgeihe: Both are still there, at least for non-PackageCompiler mode. (maybe the diff just looks strange)

CMakeLists.txt
bgeihe: Thanks for the hint!

bgeihe:
I also added the cmake call to the docs. Thanks for the reminder!

CMakeLists.txt
```diff
@@ -56,78 +56,131 @@ if( ENABLE_TESTING )
...
+    add_dependencies( ${PROJECT_NAME} PC_LIBTRIXI )
+
+    # Add linking to PackageCompiler.jl's libtrixi.so
+    target_link_libraries( ${PROJECT_NAME} INTERFACE ${PC_LIBTRIXI_SO} )
+
+    # Install configuration
+    install( DIRECTORY "${CMAKE_BINARY_DIR}/prefix-pc/lib/" TYPE LIB )
+    install( DIRECTORY "${CMAKE_BINARY_DIR}/prefix-pc/share/julia"
+             DESTINATION share
+             PATTERN "cert.pem" EXCLUDE )
```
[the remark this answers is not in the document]
sloede: It has some certificates required for communicating safely with some of the Julia servers, I believe (e.g., package servers). We should not include it, otherwise we might get hard-to-understand errors...

CMakeLists.txt
sloede: Ah yes, thanks! My bad, sorry

CMakeLists.txt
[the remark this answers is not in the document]
bgeihe: Just to clarify:

> We should not include it, otherwise we might get hard-to-understand errors...

Should we, or should we not?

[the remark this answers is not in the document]
sloede: We should not include the exclusion :-) I've already updated the code 😬

CMakeLists.txt
```diff
@@ -56,78 +56,129 @@ if( ENABLE_TESTING )
...
+    if ( ENABLE_TESTING )
+        message( NOTICE "Testing is not supported when PackageCompiler is used.")
+    endif()
 
-# Include directories, private
-target_include_directories ( ${PROJECT_NAME} PRIVATE src )
+    # Copy source with initialization routines; an object file will be created
+    file( MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/build-pc )
+    file( COPY ${CMAKE_SOURCE_DIR}/LibTrixi.jl/lib/init.c
+          DESTINATION build-pc/ )
```
sloede, line 80: I think this is not the "proper" way to do it, since it will only ever run once at configure time but not at build time. That is, the `init.c` file will be copied when CMake itself is run, but not whenever we run `make`. Thus, any changes to `init.c` *after* calling CMake will not be picked up.

I think the proper way to do this is by using a custom command to create the file, since then the output can be added as a dependency (see also [this blog post](https://crascit.com/2017/04/18/generated-sources-in-cmake-builds/)).

If I am not mistaken, something like this should work:
```suggestion
    add_custom_command( OUTPUT ${CMAKE_BINARY_DIR}/build-pc/init.c
                        COMMENT "Copying `init.c` to build folder..."
                        COMMAND ${JULIA_EXECUTABLE} -e 'cp(ARGS[1], ARGS[2])'
                                ${CMAKE_SOURCE_DIR}/LibTrixi.jl/lib/init.c
                                ${CMAKE_BINARY_DIR}/build-pc/init.c )
```

CMakeLists.txt
```diff
@@ -56,78 +56,129 @@ if( ENABLE_TESTING )
...
+
+    # Custom command to run PackageCompiler.jl to produce libtrixi.so
+    add_custom_command( OUTPUT ${PC_LIBTRIXI_SO}
+                        COMMENT "Building ${PROJECT_NAME} with PackageCompiler.jl..."
+                        COMMAND ${JULIA_EXECUTABLE}
+                                --project=${CMAKE_SOURCE_DIR}/LibTrixi.jl/lib
+                                ${CMAKE_SOURCE_DIR}/LibTrixi.jl/lib/build.jl
+                                ${JULIA_PROJECT_PATH}
+                                ${CMAKE_BINARY_DIR}/prefix-pc
+                        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/build-pc )
```
sloede, line 98:
```suggestion
                        DEPENDS ${CMAKE_BINARY_DIR}/build-pc/init.c
                        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/build-pc )
```

CMakeLists.txt
sloede: Btw, I used the Julia command to be more platform agnostic. You could of course just replace that by a `cp`...

bgeihe: Good point!

With [75344f5](https://github.com/trixi-framework/libtrixi/pull/128/commits/75344f58cb3e5ca694b0bc636a1f2f650dd6f00c) it seems to work:

- building from scratch works (as before)
- simply calling make again does not do anything (as before)
- in case `init.c` is missing in the build directory or modified in the source directory everything is rebuild

sloede: Great, thanks for checking it out!

sloede (approved):
LGTM! 🥳

PR #131, opened 2023-10-02
Fix input files for Doxygen documentation

### LINKED ISSUE #130: No overview of Fortran API?
When going to the docs, there does not seem to be a link anymore to browse the Fortran API. Has this been removed or was it just never there? And, more importantly, can we get it back?
image

sloede:
Hopefully fixes #130.


bgeihe (approved):
Thanks for the quick fix!

I just checked the sources, but apart from those listed none contain doxygen comments.

PR #137, opened 2023-10-16
Update apt-get before apt-get install

bgeihe:
Fixes #136

See https://github.com/actions/runner-images/issues/5470


sloede (approved):
Thanks!

PR #139, opened 2023-10-24
add code of conduct

ranocha:
I suggest to add the code of conduct added in this PR - mimicking what we do in Trixi.jl.


sloede (approved):
Thanks!

PR #145, opened 2023-10-27
Add build.jl as dependency for PackageCompiler

bgeihe:
If options are changed, make should re-trigger PackageCompiler


sloede (approved):
Good catch!

Issue #152, opened 2023-12-11
Julia's `Pkg` currently does not work in CI workflow


benegee:
#150 revealed an issue in the CI workflow.

When libtrixi-init-julia is invoked we get:
```
Install dependencies for configuration...
  Installing known registries into `~/.julia`
julia: symbol lookup error: /opt/hostedtoolcache/julia/1.9.4/x64/bin/../lib/julia/libcurl.so: undefined symbol: nghttp2_option_set_no_rfc9113_leading_and_trailing_ws_validation
```

- This occurs as soon as the `Pkg` package is used (instantiate, status, update, ...).
- It does not occur when `LD_PRELOAD` is emptied before. (But then we get the usual HDF5 error.)
- Since the last time the CI ran successfully, most notably julia was updated to version 1.9.4.
- Along with this, the bundled libcurl went from 4.7.0 to 4.8.0.
- The github runner has version 4.7.0 locally installed, which we load via `LD_PRELOAD: /lib/x86_64-linux-gnu/libcurl.so.4`.
- The new bundled libcurl version has several more `nghttp2_*` symbols.
- When `Pkg` does something, Julia loads the bundled libcurl.so using dlopen. This happens independent of any previously loaded libcurl.
- At this point both libcurl.so seem to be loaded
  ```
  julia> Libdl.dllist()
  74-element Vector{String}:
  "linux-vdso.so.1"
  "/usr/lib/x86_64-linux-gnu/libcurl.so"
  "/lib/x86_64-linux-gnu/libnghttp2.so.14"
  [...]
  "/mnt/ssd/home/bgeihe/julia-1.9.4/bin/../lib/julia/libcurl.so"
  ```

- The bundled libcurl contains `nghttp2_*` symbols which are not provided by the system libnghttp2

PR #156, opened 2024-01-08
Update to julia 1.10

bgeihe:


bgeihe:
Fortunately fixes #152 

But has some issue with PackageCompiler
```
- PackageCompiler: compiling fresh sysimage (incremental=false)
error during bootstrap:
LoadError("/tmp/jl_HeuRo1/sysimage_packagecompiler_47e340e8-ae21-11ee-2456-a7170323e6fe.jl", 18, LoadError("/opt/hostedtoolcache/julia/1.10.0/x64/share/julia/stdlib/v1.10/Test/src/Test.jl", 3, LoadError("/opt/hostedtoolcache/julia/1.10.0/x64/share/julia/stdlib/v1.10/Test/src/precompile.jl", 1, MethodError(Base.RedirectStdStream(1, true), (Core.CoreSTDOUT(),), 0x0000000000008634))))
```

bgeihe:
So, we could matrix-test with Julia versions 1.9.3 (which does not fall victim to #152) and with 1.10.*. Coverage would only be tested with either of them. Still it would be nice to have PackageCompiler working with 1.10. Any ideas @sloede ?

bgeihe:
I would like to have green lights on main again. So, for the time being, I would like to use julia 1.10 as default, and 1.9.3 for test package compiler.

sloede (changes requested):
Instead of guarding all these PC.jl runs against usage with v1.10 in the if conditions, why don't you just remove the `package-compiler` from the test matrix in line 35 (https://github.com/trixi-framework/libtrixi/pull/156/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR35)? Shouldn't this then do exactly what you want: run the default items in the test matrix _except_ for the PC.jl stuff, which you explicitly add for Julia v1.9.3 in the `include` section. Or am I missing something?

bgeihe:
Yeah, that is way more clever!

sloede (approved):

Issue #43, opened 2023-07-05
Initial API for data transfers between main program and libtrixi


sloede:
Here are some initial thoughts for an API that would allow one to do more than just control the main time integration loop from outside. The idea is to keep this initial post updated (by anyone) with the results of the subsequent discussion.

### Basic querying
Functions to get basic information from Trixi.jl.
```c
int trixi_ndims(int handle);         // Return number of spatial dimensions
int trixi_nelements(int handle);     // Return number of elements (cells)
int trixi_polydeg(int handle);       // Return polynomial degree of DGSEM approximation
int trixi_nvariables(int handle);    // Return number of (conservative) variables
int trixi_ndofs(int handle);         // Return total number of degrees of freedom
int trixi_ndofs_element(int handle); // Return number of degrees of freedom for one element
```

### Raw data exchange
Direct access to solution data `u`.
```c
void trixi_load_u(double* u, int handle);        // Load data from Trixi.jl into `u`
void trixi_store_u(const double* u, int handle); // Store data from `u` in Trixi.jl
```

### Block-structured data exchange
DGSEM has >> 1 degrees of freedom (DOF) per element. However, other programs might not care for our Gauss node distribution and just want equidistant data (a.k.a. finite-volume-type block-structured data), e.g., 8x8x8 cell-centered values per element. Thus we provide convenience functions that allow loading and storing in such block-structured format and do the conversion to Gauss nodes internally.
```c
void trixi_set_blocksize(int blocksize, int handle);   // Set 1D block size
int  trixi_get_blocksize(int handle);                  // Get 1D block size
int  trixi_ndofs_block(int handle);                    // Return total #DOFs for block-structured data
void trixi_load_u_block(double* u, int handle);        // Load block-structured data from Trixi.jl
void trixi_store_u_block(const double* u, int handle); // Store block structured data in Trixi.jl
```

PR #169, opened 2024-02-22
Extend data access API

benegee:
### New
- `trixi_ndofs(int handle);`
- `trixi_ndofsglobal(int handle);`
- `trixi_ndofselement(int handle);`
- `trixi_nnodes(int handle);`
- `trixi_load_primitive_vars(int handle, int variable_id, double * data);`
- `trixi_load_node_reference_coordinates(int handle, double* node_coords);`
- `trixi_load_node_weights(int handle, double* node_weights);`

### Breaking changes
- `trixi_nelementsglobal(int handle);` (was `trixi_nelements_global(int handle);`)
- `trixi_load_element_averaged_primitive_vars(int handle, int variable_id, double * data);` (was `trixi_load_cell_averages(double * data, int handle);`

### Conventions
Adopt https://trixi-framework.github.io/Trixi.jl/stable/conventions/#Cells-vs.-elements-vs.-nodes. For libtrixi, avoid "cell" in favor of "element".

Related #43


sloede (commented):
Looks sensible to me. Just note that changing the public API requires a breaking release. That's OK, of course.

Alternatively, you could consider to only *extend* the API, e.g., by adding a new `trixi_load_cell_average` or `trixi_load_cell_averages_single` function. That has the advantage that you can still decide to introduce a breaking change later. It has the downside, that you need more tests.

LibTrixi.jl/src/LibTrixi.jl
```diff
@@ -31,12 +31,24 @@ export trixi_nelements,
...
        trixi_nelements_global_jl
+export trixi_ndofs,
+       trixi_ndofs_cfptr,
+       trixi_ndofs_jl
+export trixi_ndofs_global,
+       trixi_ndofs_global_cfptr,
+       trixi_ndofs_global_jl
+export trixi_ndofs_element,
+       trixi_ndofs_element_cfptr,
+       trixi_ndofs_element_jl
```
sloede, line 42: I understand the need to export these functions. Have you considered whether it might be more helpful to call them `trixi_ndofs`, `trixi_ndofsglobal`, and `trixi_ndofselement` instead of using `_` in the middle? Since they names are _very_ similar to the original functions used in Trixi.jl, I fear that it might lead to confusion that sometimes it's with an underscore and sometimes it isn't.

LibTrixi.jl/src/LibTrixi.jl
```diff
@@ -31,12 +31,24 @@ export trixi_nelements,
...
 export trixi_nvariables,
        trixi_nvariables_cfptr,
        trixi_nvariables_jl
 export trixi_load_cell_averages,
        trixi_load_cell_averages_cfptr,
        trixi_load_cell_averages_jl
+export trixi_load_prim,
+       trixi_load_prim_cfptr,
+       trixi_load_prim_jl
```
sloede, line 51: I am not a fan of arbitrary abbreviations unless there is a good reason for it. `cons2prim` and friends are named this way for historic reasons, and since they require a certain brevity. Here, on the other hand, I'd recommend to use something more descriptive, e.g., `trixi_load_primitive_variables` (or `_vars`, since that should be descriptive enough).

LibTrixi.jl/src/LibTrixi.jl
sloede, line 42: Below, we have `trixi_load_cell_averages` and friends. I think it would be prudent to decide whether to call it `elements` or `cells` in the context of libtrixi, similar to what we did for Trixi.jl.

If you decide to not make a distinction between cells and elements in libtrixi, it might make things easier for libtrixi users, but harder for Trixi.jl users. However, in that case I recommend to pick one (either "cells" or "elements") and use it consistently everywhere, but to not mix and match. If you want to continue to use both like in Trixi.jl, I recommend changing "trixi_load_cell_averages" accordingly (not in this PR, but in a future one)

sloede: Note that this distinction (or not) should also be consequently used in all comments, where, I believe, also cells and elements are used currently.

LibTrixi.jl/src/api_c.jl
```diff
@@ -312,6 +312,51 @@ end
...
+    return trixi_ndofs_jl(simstate)
+end
+
+trixi_ndofs_cfptr() = @cfunction(trixi_ndofs, Cint, (Cint,))
+
+
+"""
+    trixi_ndofs_global(simstate_handle::Cint)::Cint
+
+Return number of global eldegrees of freedom (all quadrature points on all cells).
```
sloede, line 333:
```suggestion
Return global number of degrees of freedom (all quadrature points on all cells on all ranks).
```

LibTrixi.jl/src/api_c.jl
```diff
@@ -328,32 +373,60 @@ trixi_nvariables_cfptr() = @cfunction(trixi_nvariables, Cint, (Cint,))
 
 
 """
-    trixi_load_cell_averages(data::Ptr{Cdouble}, simstate_handle::Cint)::Cvoid
+    trixi_load_cell_averages(data::Ptr{Cdouble}, index::Cint, simstate_handle::Cint)::Cvoid
```
sloede, line 376: As mentioned before, remember to make a breaking release after this change. It might be good to think whether other breaking changes can be included as well before the next release, e.g., regarding the cell/element naming choice.

LibTrixi.jl/src/api_c.jl
```diff
@@ -328,32 +373,60 @@ trixi_nvariables_cfptr() = @cfunction(trixi_nvariables, Cint, (Cint,))
...
 
 Return cell averaged solution state.
```
sloede, line 378: I hadn't noticed before, but I think "Return" is the wrong word here, since the function does not return anything...

LibTrixi.jl/src/api_c.jl
```diff
@@ -328,32 +373,60 @@ trixi_nvariables_cfptr() = @cfunction(trixi_nvariables, Cint, (Cint,))
...
 
-Cell averaged values for each cell and each primitive variable are stored in a contiguous
-array, where cell values for the first variable appear first and values for the other
-variables subsequently (structure-of-arrays layout).
+Cell averaged values for the primitive variable at position `index` for each cell are stored
+in the given array `data`.
 
 The given array has to be of correct size and memory has to be allocated beforehand.
```
sloede, line 383: Maybe give a hint on how to determine the correct size that needs to be allocated?

LibTrixi.jl/src/api_c.jl
```diff
@@ -328,32 +373,60 @@ trixi_nvariables_cfptr() = @cfunction(trixi_nvariables, Cint, (Cint,))
...
 
 trixi_load_cell_averages_cfptr() =
-    @cfunction(trixi_load_cell_averages, Cvoid, (Ptr{Cdouble}, Cint,))
+    @cfunction(trixi_load_cell_averages, Cvoid, (Ptr{Cdouble}, Cint, Cint,))
+
+
+"""
+    trixi_load_prim(data::Ptr{Cdouble}, index::Cint, simstate_handle::Cint)::Cvoid
+
+Return primitive variable.
```
sloede, line 406: As above, "return" is imho the wrong word here

LibTrixi.jl/src/api_c.jl
```diff
@@ -328,32 +373,60 @@ trixi_nvariables_cfptr() = @cfunction(trixi_nvariables, Cint, (Cint,))
...
+
+The values for the primitive variable at position `index` at every degree of freedom for
+the simulation given by `simstate_handle` is stored in the given array `data`.
+
+The given array has to be of correct size and memory has to be allocated beforehand.
```
sloede, line 411: As above

LibTrixi.jl/src/api_c.jl
sloede, line 376:
```suggestion
    trixi_load_cell_averages(data::Ptr{Cdouble}, variable_id::Cint, simstate_handle::Cint)::Cvoid
```
`index` is super generic, I suggest to use something more specific (I think `variable_id` or similar is what we use in Trixi.jl). Also, it might be good to add a comment to the docstring on how to obtain the minimum and maximum variable_id, and to communicate the fact that this is one-based indexing.

src/api.c
```diff
@@ -488,6 +496,65 @@ int trixi_nelements_global(int handle) {
...
+ *
+ * @brief Return number of global degrees of freedom.
+ *
+ * These usually differ from the local count when doing parallel computations.
+ *
+ * @param[in]  handle  simulation handle
+ *
+ * @see trixi_ndofs_api_c
+ */
+int trixi_ndofs_global(int handle) {
```
sloede, line 531: Most of my comments re naming conventions etc. that I left above also apply to the C/Fortran API, so I won't repeat them here

sloede (changes requested):
Looks mostly good to me, but I think there are some structural decisions to be made (or maybe they have been made but I am just not aware 😬)

LibTrixi.jl/src/LibTrixi.jl
bgeihe: Should it rather be `trixi_ndofselement` or `trixi_ndofscell`? (something like this does not exist in Trixi)

bgeihe:
Say I want to stay as close to Trixi.jl as possible and have read
https://trixi-framework.github.io/Trixi.jl/stable/conventions/#Cells-vs.-elements-vs.-nodes
would there be any place in libtrixi where "cell" should be used?

bgeihe:
Great! Thanks for your hints!

This is now totally cell-free. And totally breaking.

benegee:
Thanks for reviewing @sloede ! 
PackageCompiler is still failing. I had to discard my plan to resolve this via #207 because there seem to be some new issues.
It would however be great to merge this anyway.

sloede (approved):

PR #173, opened 2024-02-29
RFC: source terms via global vector storage

benegee:
This demonstrates how source terms can be evaluated via a global shared storage.

Issues:
- Data vectors need to be access via an index. Trixi.jl's `source_terms` function works with the local coordinate `x`. Therefore a mapping `x -> index` is required or a change in Trixi.jl to allow for a `source_terms` function operating on indices.
- In this demonstration also a vector of node coordinates is fetched. This would become obsolete if a (t8code) mesh existed both in the calling program and within libtrixi.

Some rough timing results:

<details>
<summary>Timing libtrixi</summary>

```
────────────────────────────────────────────────────────────────────────────────────────────────────
Trixi.jl simulation finished.  Final time: 2.0  Time steps: 229 (accepted), 229 (total)
────────────────────────────────────────────────────────────────────────────────────────────────────


*** Trixi controller ***   Finalize Trixi simulation
 ────────────────────────────────────────────────────────────────────────────────────
              Trixi.jl                      Time                    Allocations
                                   ───────────────────────   ────────────────────────
         Tot / % measured:              2.45s /  16.8%            137MiB /  13.5%

 Section                   ncalls     time    %tot     avg     alloc    %tot      avg
 ────────────────────────────────────────────────────────────────────────────────────
 I/O                            5    199ms   48.4%  39.7ms   10.6MiB   57.4%  2.11MiB
   save solution                4    123ms   29.9%  30.7ms   10.3MiB   56.1%  2.59MiB
   ~I/O~                        5   76.0ms   18.5%  15.2ms    230KiB    1.2%  46.1KiB
   save mesh                    4    664ns    0.0%   166ns     0.00B    0.0%    0.00B
   get element variables        4    415ns    0.0%   104ns     0.00B    0.0%    0.00B
   get node variables           4   87.0ns    0.0%  21.8ns     0.00B    0.0%    0.00B
 rhs!                       1.15k    119ms   29.0%   104μs   5.14KiB    0.0%    4.59B
   volume integral          1.15k   48.2ms   11.7%  42.1μs     0.00B    0.0%    0.00B
   interface flux           1.15k   30.9ms    7.5%  27.0μs     0.00B    0.0%    0.00B
   source terms             1.15k   21.2ms    5.2%  18.5μs     0.00B    0.0%    0.00B
   surface integral         1.15k   11.5ms    2.8%  10.1μs     0.00B    0.0%    0.00B
   Jacobian                 1.15k   4.40ms    1.1%  3.84μs     0.00B    0.0%    0.00B
   reset ∂u/∂t              1.15k   2.29ms    0.6%  2.00μs     0.00B    0.0%    0.00B
   ~rhs!~                   1.15k    563μs    0.1%   491ns   5.14KiB    0.0%    4.59B
   boundary flux            1.15k   22.1μs    0.0%  19.3ns     0.00B    0.0%    0.00B
 analyze solution               4   88.4ms   21.5%  22.1ms   7.85MiB   42.6%  1.96MiB
 calculate dt                 230   4.59ms    1.1%  19.9μs     0.00B    0.0%    0.00B
 ────────────────────────────────────────────────────────────────────────────────────
```
</details>

<details>
<summary>Timing Trixi.jl</summary>

```
────────────────────────────────────────────────────────────────────────────────────
              Trixi.jl                      Time                    Allocations
                                   ───────────────────────   ────────────────────────
         Tot / % measured:              164ms /  94.0%           2.00MiB /  90.6%

 Section                   ncalls     time    %tot     avg     alloc    %tot      avg
 ────────────────────────────────────────────────────────────────────────────────────
 rhs!                       1.15k    144ms   93.2%   126μs   5.14KiB    0.3%    4.59B
   source terms             1.15k   59.7ms   38.7%  52.1μs     0.00B    0.0%    0.00B
   volume integral          1.15k   41.5ms   26.9%  36.2μs     0.00B    0.0%    0.00B
   interface flux           1.15k   26.6ms   17.3%  23.2μs     0.00B    0.0%    0.00B
   surface integral         1.15k   9.68ms    6.3%  8.44μs     0.00B    0.0%    0.00B
   Jacobian                 1.15k   3.82ms    2.5%  3.34μs     0.00B    0.0%    0.00B
   reset ∂u/∂t              1.15k   1.99ms    1.3%  1.73μs     0.00B    0.0%    0.00B
   ~rhs!~                   1.15k    498μs    0.3%   434ns   5.14KiB    0.3%    4.59B
   boundary flux            1.15k   28.8μs    0.0%  25.2ns     0.00B    0.0%    0.00B
 I/O                            5   4.82ms    3.1%   964μs   1.56MiB   85.9%   319KiB
   save solution                4   4.31ms    2.8%  1.08ms   1.55MiB   85.4%   397KiB
   ~I/O~                        5    507μs    0.3%   101μs   8.83KiB    0.5%  1.77KiB
   get element variables        4    765ns    0.0%   191ns     0.00B    0.0%    0.00B
   save mesh                    4    488ns    0.0%   122ns     0.00B    0.0%    0.00B
   get node variables           4    342ns    0.0%  85.5ns     0.00B    0.0%    0.00B
 calculate dt                 230   4.04ms    2.6%  17.6μs     0.00B    0.0%    0.00B
 analyze solution               4   1.59ms    1.0%   396μs    258KiB   13.9%  64.5KiB
 ────────────────────────────────────────────────────────────────────────────────────
```
</details>

Confer #172


bgeihe:
Fortran interface is still missing.

benegee:
Ready for review @sloede 
A real example using this feature will be provided by #177

LibTrixi.jl/src/api_c.jl
```diff
@@ -461,6 +461,53 @@ trixi_load_primitive_vars_cfptr() =
...
+end
+
+trixi_store_in_database_cfptr() =
+    @cfunction(trixi_store_in_database, Cvoid, (Cint, Cint, Cint, Ptr{Cdouble},))
+
+
+"""
+    trixi_get_time(simstate_handle::Cint)::Cdouble
+
+Return current physical time.
```
sloede, line 499: should we call this "simulation_time" or something similar to be clear what this means?

LibTrixi.jl/src/api_c.jl
```diff
@@ -461,6 +461,53 @@ trixi_load_primitive_vars_cfptr() =
...
+Store data vector in current simulation's database.
+
+A reference to the passed data array `data` will be stored in the database of the simulation
+given by `simstate_handle` at given `index`. The database object has to be created in
+`init_simstate()` of the running libelixir and can be used throughout the simulation.
+
+The database object has to exist, has to be of type `LibTrixiDataBaseType`, and has to hold
+enough data references such that access at `index` is valid.
+
+The size of `data` has to match `size`.
```
sloede, line 477: IMHO it would be good to emphasize here that the memory storage remains on the user side and that whenever the underlying object is freed, it should also be removed from the database

LibTrixi.jl/src/api_c.jl
```diff
@@ -461,6 +461,53 @@ trixi_load_primitive_vars_cfptr() =
     @cfunction(trixi_load_primitive_vars, Cvoid, (Cint, Cint, Ptr{Cdouble}))
 
 
+"""
+    trixi_store_in_database(data::Ptr{Cdouble}, size::Cint, index::Cint,
+                            simstate_handle::Cint)::Cvoid
```
sloede, line 466: I am not 100% sure I like the name "database", since it has a very specific meaning in computer science. Also, it does not really "store" any data but only a reference to it, right? To me, the current name implies that data is actually saved into some Trixi-allocated containers.

Alas, I do not have a better name, but maybe you have another idea?

sloede, line 466: Why does the user have to provide the `index`? Wouldn't it be nicer if one could just dump data into the database (or container (or data storage (or data container (or data lake)))) and retrieve the index from the function, just like we do with simstate?

LibTrixi.jl/src/simulationstate.jl
```diff
@@ -1,12 +1,21 @@
+const LibTrixiDataBaseType = Vector{Ref{Vector{Float64}}}
```
sloede, line 1: I wouldn't suffix a type with `Type` :-) Why not just `LibTrixiData{Base,Store,Whatever}`?

sloede (changes requested):

LibTrixi.jl/src/api_c.jl
benegee: What about registry?

LibTrixi.jl/src/api_jl.jl
```diff
@@ -199,6 +199,20 @@ function trixi_load_element_averaged_primitive_vars_jl(simstate, variable_id, da
 end
 
 
+function trixi_store_in_database_jl(simstate, index, data)
+    simstate.data[index] = Ref(data)
```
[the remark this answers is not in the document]
benegee: Thanks for the hint!
I somehow must have assumed that assigning a vector would deep-copy its memory.

LibTrixi.jl/src/api_c.jl
benegee: Done!

LibTrixi.jl/src/api_c.jl
benegee: This is related to a general flaw of the approach: 

- The data registry has to be created in `init_simstate` of the libelixir, which is called when the simulation is initialized
- `init_simstate` has to be called first in order to get a simulation handle
- Only then, data vectors can be registered
- However, during initialization of the simulation, Trixi will already call calc_sources, where the data is accessed
- To make this work, some dummy data vector has to be added during initialization and has to be overwritten later

LibTrixi.jl/src/simulationstate.jl
benegee: Done. Now called `LibTrixiDataRegistry`.

LibTrixi.jl/src/api_c.jl
sloede: Why not? Although registry has a specific meaning in Julia as well... what about "data registry" (too long?) or "data bridge" (to emphasize its purpose) or "data pool" (like cache, but cache is already with a strong meaning in the Trixi.jl world)?

sloede: OK. This sounds really not like an optimal solution. But it's the one we currently ahve, so let's see how far we can take it 😅.

LibTrixi.jl/src/simulationstate.jl
sloede: Nice!

LibTrixi.jl/src/api_c.jl
sloede: I just saw your comment that you renamed it to "data registry", so that's that 😊

sloede (approved):

PR #177, opened 2024-03-04
Add baroclinic instability test case

bgeihe:
Just the libelixir.


LibTrixi.jl/examples/libelixir_p4est3d_euler_baroclinic_instability.jl
```diff
@@ -0,0 +1,281 @@
+# An idealized baroclinic instability test case
+#
```
sloede, line 2:
```suggestion
#
# Note that this libelixir is based on the original baroclinic instability elixir by
# Erik Faulhaber for Trixi.jl
# Source: https://github.com/trixi-framework/Trixi.jl/blob/main/examples/p4est_3d_dgsem/elixir_euler_baroclinic_instability.jl
#
```

sloede (dismissed):
Maybe add the reference to the original, but other than that, LGTM

PR #192, opened 2024-06-13
Swap installation order of Trixi.jl and LibTrixi.jl

### LINKED ISSUE #190: LibTrixi.jl does not pull in latest Trixi.jl version
When setting up libtrixi using `libtrixi-init-julia`, not the latest Trixi.jl version is installed, but v0.5.36. See e.g. https://github.com/trixi-framework/libtrixi/actions/runs/8964091186/job/24615337557#step:16:805

In the` [compat]` section we have
`Trixi = "0.5.29, 0.6, 0.7"`
and indeed Trixi can be upgraded afterwards without problems.  

I tried to reproduce this with a LibTrixi.jl dummy package, which only has Trixi.jl as a dependency. I started with a fresh Julia project and `]dev <path>/LibTrixi.jl`, which resulted in
```
Status `~/install/libtrixi-julia-test/Manifest.toml`
...
⌃ [a7f1ee26] Trixi v0.6.5 (<v0.7.13)
```
(The same happens with the real LibTrixi.jl package, i.e. `]add LibTrixi`)

As indicated, Trixi.jl can be updated. However `DiffEqBase` is then downgraded:
```
⌅ [2b5f629d] ↓ DiffEqBase v6.145.2 ⇒ v6.143.0
```

Is this the reason why Trixi v0.6.5 was chosen in the first place? (The `DiffEqBase` compat entry for Trixi.jl was introduced in v0.6.6)

Starting again with  a fresh project and doing `pkg> add Trixi` in the first place directly installs ` [a7f1ee26] + Trixi v0.7.13`, and `]dev <path>/LibTrixi.jl` afterwards succeeds without downgrading Trixi.jl.
So, should we just swap the order of installation?

benegee:
Does this fix #190 ?

### DIFF
--- a/LibTrixi.jl/src/api_jl.jl
+++ b/LibTrixi.jl/src/api_jl.jl
@@ -77,8 +77,8 @@ end
 
 
 function trixi_nelements_global_jl(simstate)
-    _, _, solver, cache = mesh_equations_solver_cache(simstate.semi)
-    return nelementsglobal(solver, cache)
+    mesh, _, solver, cache = mesh_equations_solver_cache(simstate.semi)
+    return nelementsglobal(mesh, solver, cache)
 end
 
 
--- a/utils/libtrixi-init-julia
+++ b/utils/libtrixi-init-julia
@@ -335,7 +335,10 @@ fi
 [ $? -eq 0 ] || die "could not configure system HDF5 library for Julia"
 echo
 
-# Install remaining dependencies
+# Install Trixi.jl and OrdinaryDiffEq.jl first
+# Rationale: By first installing Trixi.jl, we ensure that we use its latest version,
+# even if it does not play nicely with the latest version of OrdinaryDiffEq.jl.
+# xref: https://github.com/trixi-framework/libtrixi/issues/190
 echo "Install Trixi.jl and OrdinaryDiffEq.jl..."
 JULIA_DEPOT_PATH="$julia_depot" JULIA_PKG_PRECOMPILE_AUTO=0 \
     $julia_exec --project=. \
--- a/utils/libtrixi-init-julia
+++ b/utils/libtrixi-init-julia
@@ -335,14 +335,6 @@ fi
 [ $? -eq 0 ] || die "could not configure system HDF5 library for Julia"
 echo
 
-# Develop LibTrixi.jl
-echo "Install LibTrixi.jl..."
-JULIA_DEPOT_PATH="$julia_depot" JULIA_PKG_PRECOMPILE_AUTO=0 \
-    $julia_exec --project=. \
-    -e "using Pkg; Pkg.develop(path=\"$libtrixi_jl_path\")"
-[ $? -eq 0 ] || die "could not install LibTrixi.jl"
-echo
-
 # Install remaining dependencies
 echo "Install Trixi.jl and OrdinaryDiffEq.jl..."
 JULIA_DEPOT_PATH="$julia_depot" JULIA_PKG_PRECOMPILE_AUTO=0 \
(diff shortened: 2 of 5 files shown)


benegee:
I also updated t8code's version to v2.0.0. Should have done that earlier, but it also seems to be necessary with the current version of Trixi.jl.

utils/libtrixi-init-julia
```diff
@@ -335,14 +335,6 @@ fi
 [ $? -eq 0 ] || die "could not configure system HDF5 library for Julia"
 echo
 
-# Develop LibTrixi.jl
-echo "Install LibTrixi.jl..."
-JULIA_DEPOT_PATH="$julia_depot" JULIA_PKG_PRECOMPILE_AUTO=0 \
-    $julia_exec --project=. \
-    -e "using Pkg; Pkg.develop(path=\"$libtrixi_jl_path\")"
-[ $? -eq 0 ] || die "could not install LibTrixi.jl"
-echo
```
sloede, line 344: Is this change even necessary now that you use the newer t8code version?

If yes, I recommend to expand the comment in the line below.

utils/libtrixi-init-julia
```diff
@@ -335,14 +335,6 @@ fi
...
-
 # Install remaining dependencies
```
sloede, line 338:
```suggestion
# Install Trixi.jl and OrdinaryDiffEq.jl first
# Rationale: By first installing Trixi.jl, we ensure that we use its latest version,
# even if it does not play nicely with the latest version of OrdinaryDiffEq.jl.
# xref: https://github.com/trixi-framework/libtrixi/issues/190
```

sloede (changes requested):

utils/libtrixi-init-julia
benegee: :thinking: 

I would argue that at his point during setup the _external_ t8code library does not play a role, yet.

But apart from that: This segfault made me try a different t8code version
https://github.com/trixi-framework/libtrixi/actions/runs/9503077644/job/26192654449#step:22:903

Now I am thinking about compatibility of T8code.jl and the external t8code library: Trixi.jl specifies compat bounds for T8code.jl, but how about T8code.jl and t8code when not using jlls?

sloede: When using external libraries, there's just no way to specify compat bounds. We can - and do - require compat bounds on the JLL, see
https://github.com/DLR-AMR/T8code.jl/blob/5c9b43f95c703df8bb06f90c683717c64b42b985/Project.toml#L25
Other than that, I think it's up to the user to do the right thing...

benegee: Good point! This gives an idea on which version to use.

So, what's missing here?

sloede: > So, what's missing here?

What do you mean by that?

benegee: Sorry!

> Is this change even necessary now that you use the newer t8code version?
Yes.

> If yes, I recommend to expand the comment in the line below.
Do you think we should add anything else in this PR (besides your explaining comment)?

sloede: Nope, all good then, I guess 😊

sloede (approved):

sloede:
> 🙄 [trixi-framework/Trixi.jl#1865](https://github.com/trixi-framework/Trixi.jl/pull/1865) (which happens to be my own PR) changed the non-public-API `nelementsglobal`.
> 
> What to do? Restrict Trixi.jl version to >=0.7.16?

What do you mean? All tests seem to pass?!

LibTrixi.jl/src/api_jl.jl
```diff
@@ -77,8 +77,8 @@ end
 
 
 function trixi_nelements_global_jl(simstate)
-    _, _, solver, cache = mesh_equations_solver_cache(simstate.semi)
-    return nelementsglobal(solver, cache)
+    mesh, _, solver, cache = mesh_equations_solver_cache(simstate.semi)
+    return nelementsglobal(mesh, solver, cache)
```
benegee, line 81: This is required when using Trixi.jl v0.7.16 or above.

sloede: Yes, I guess you need to bump the minimum version then 👍

sloede (approved):

PR #215, opened 2024-11-11
Move to new t8code version 3.0.0

benegee:
- Move to latest t8code version
- Use t8code's cmake support
- Enable new t8code Fortran interface 
- Add t8code Fortran controller example
- Adapt the docs


benegee:
PackageCompiler build currently fails but will be fixed in course of https://github.com/trixi-framework/libtrixi/pull/207 hereafter.

benegee:
Another ping @sloede as requested. :smiley:

CMakeLists.txt
```diff
@@ -36,7 +38,7 @@ list ( APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/" )
 find_package( Julia REQUIRED )
 
 # Find t8code
-find_package( T8code )
+find_package( T8CODE CONFIG )
```
sloede, line 41: Why `CONFIG`?

LibTrixi.jl/src/api_jl.jl
```diff
@@ -46,6 +46,12 @@ function trixi_finalize_simulation_jl(simstate)
         end
     end
 
+    # finalize T8codeMesh explicitly
```
sloede, line 49: Why is this required? That is, why not let the finalizer of the Julia object do it for you once it's released by GC?

sloede: If it is needed e.g. to give back all the memory sooner, I suggest to leave this as a comment in the code. You can then just resolve this convo

test/c/CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ foreach ( TEST ${TESTS} )
         PRIVATE ${CMAKE_SOURCE_DIR}/src
     )
     if ( T8CODE_FOUND )
-        target_include_directories( ${TARGET_NAME} PRIVATE ${T8CODE_INCLUDE_DIR} )
+        target_include_directories( ${TARGET_NAME} PRIVATE ${T8CODE_ROOT}/include/t8_fortran_interface/ )
```
sloede, line 35: Why only include the fortran interface dir?

sloede: Especially given that this is in the `c` directory

sloede (commented):
Generally LGTM, just a few questions. Also, some tests seem to be failing?

CMakeLists.txt
benegee: Short answer: you get annoying warnings else

Long answer: there are two modes in which `find_package` operates: https://cmake.org/cmake/help/latest/command/find_package.html#search-modes
"module mode" is the default one and will first try to find a `FindT8code.cmake` file. T8code however provides `T8CODEConfig.cmake` and this is only found by "config mode".

LibTrixi.jl/src/api_jl.jl
benegee: This is related to the latest changes in Trixi for t8code, see also https://github.com/trixi-framework/Trixi.jl/pull/1980#discussion_r1830524774

Currently `finalize` has to be called, otherwise an error is thrown. We do not have something like an exit hook feature in libtrixi, so you cannot call it from the libelixir.

test/c/CMakeLists.txt
benegee: Good catch, thanks!

With the new t8code cmake module, `target_link_libraries( ${TARGET_NAME} PRIVATE T8CODE::T8 )` magically seems to also add the corresponding include directories. The Fortran module however sits in a subdirectory, so I had to add this directory manually.

Anyways  it should not be needed here for the C tests.

benegee: Verified, still works.

LibTrixi.jl/src/api_jl.jl
sloede: OK. Do you maybe want to add a comment for this?

CMakeLists.txt
sloede: OK, thanks. Is this maybe an issue that should be reported upstream, since it seems to be non-standard and confusing (at least to me 😬).

LibTrixi.jl/src/api_jl.jl
jmark: This is an un-feature of t8code interacting with the Julia GC. See this discussion here: https://github.com/DLR-AMR/t8code/issues/1295.

I'm thinking about a solution for that. But in the meantime finalizing explicitly is the best workaround.

CMakeLists.txt
jmark: Actually not t8code's fault. The CMake documentation is pretty clear on that:
> The Find<PackageName>.cmake file is not typically provided by the package itself. Rather, it is normally provided by something external to the package, such as the operating system, CMake itself, or even the project from which the find_package() command was called. Being externally provided, [Find Modules](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#find-modules) tend to be heuristic in nature and are susceptible to becoming out-of-date. They typically search for certain libraries, files and other package artifacts.

So, t8code does not support `module` mode per default. This is up to the system admins offering t8code as a module.

jmark: So, I would say, `find_package( T8CODE CONFIG )` is the way to go to avoid any superfluous warnings.

LibTrixi.jl/src/api_jl.jl
benegee: Comment added.

sloede: Thanks a lot!

CMakeLists.txt
sloede: TIL! Thanks for the explanation!

LibTrixi.jl/src/api_jl.jl
```diff
@@ -46,6 +46,14 @@ function trixi_finalize_simulation_jl(simstate)
         end
     end
 
+    # In course of garbage collection, MPI might get finalized before t8code related objects.
+    # This can lead to crashes because t8code allocates MPI related objects, e.g. shared
+    # memory arrays. The workaround is to finalize T8codeMesh explicitly in advance.
```
sloede, line 51:
```suggestion
    # In course of garbage collection, MPI might get finalized before t8code related objects.
    # This can lead to crashes because t8code allocates MPI related objects, e.g. shared
    # memory arrays. The workaround is to finalize T8codeMesh explicitly in advance.
    # x-ref: https://github.com/DLR-AMR/t8code/issues/1295
    # x-ref: https://github.com/trixi-framework/libtrixi/pull/215#discussion_r1843676330
```

sloede (approved):
Thanks!

Issue #178, opened 2024-03-06
Failing precompilation of libtrixi


bgeihe:
Currently precompilation of libtrixi does not work on rocinante when starting from scratch. The error is:

```
Failed to precompile LibTrixi [7e097bd5-a775-4bc1-90b6-ad92bd7220c1] to "./julia-depot/compiled/v1.10/LibTrixi/jl_oEM50J".

[1404813] signal (11.1): Segmentation fault
in expression starting at none:0
opal_hwloc_base_free_topology at /lib/x86_64-linux-gnu/libopen-pal.so.40 (unknown line)
unknown function (ip: 0x7fffdb832d16)
mca_base_framework_close at /lib/x86_64-linux-gnu/libopen-pal.so.40 (unknown line)
opal_finalize at /lib/x86_64-linux-gnu/libopen-pal.so.40 (unknown line)
ompi_mpi_finalize at /lib/x86_64-linux-gnu/libmpi.so (unknown line)
MPI_Finalize at /mnt/ssd/home/bgeihe/MPI_Finalize_MWE/julia-depot/packages/MPI/z2owj/src/api/generated_api.jl:1872 [inlined]
Finalize at /mnt/ssd/home/bgeihe/MPI_Finalize_MWE/julia-depot/packages/MPI/z2owj/src/environment.jl:263
#7 at /mnt/ssd/home/bgeihe/MPI_Finalize_MWE/julia-depot/packages/MPI/z2owj/src/environment.jl:140
unknown function (ip: 0x7fffe0ded907)
_jl_invoke at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/gf.c:2894 [inlined]
ijl_apply_generic at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/gf.c:3076
_atexit at ./initdefs.jl:426
jfptr__atexit_79296.1 at /mnt/hd1/opt/julia/1.10.0/lib/julia/sys.so (unknown line)
_jl_invoke at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/gf.c:2894 [inlined]
ijl_apply_generic at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/gf.c:3076
jl_apply at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/julia.h:1982 [inlined]
[...]

PR #241, opened 2025-06-02
(Temporary) workaround for failing `HDF5_jll` precomilation

benegee:
Given the fact that `HDF5_jll` is not used for our actual tests, we can ignore the failing precompilation in the setup steps. This would allow us to have our CI running again until there is a proper fix for #178, maybe via https://github.com/JuliaPackaging/Yggdrasil/pull/10977.


benegee:
Thanks for adding the comments!

PR #252, opened 2025-08-25
Fix package compiler and regular builds

### LINKED ISSUE #253: PackageCompiler failing: `no method matching rand(::Int64)`
Seems to be caused by recent changes in PreallocationTools.jl, where `rand` is now used.
https://github.com/SciML/PreallocationTools.jl/commit/b0c743c279c1d0c4ef2e2a51dca3ca9292007444

benegee:
~~Add `PreallocationTools = "< 0.4.30"` compat.~~

Use `filter_stdlibs = false` (thanks to @vchuravy )
(we found in #216 that the resulting shared object is large anyways)
Resolves #253 

Use `Trixi.False() ` instead of `OrdinaryDiffEq.False()`.

### DIFF
--- a/LibTrixi.jl/lib/build.jl
+++ b/LibTrixi.jl/lib/build.jl
@@ -43,7 +43,7 @@ lib_name = "trixi"
 incremental = false
 
 # Do not include stdlibs which are not needed
-filter_stdlibs = true
+filter_stdlibs = false
 
 # Overwrite existing files/folders at `dest_dir`
 force = true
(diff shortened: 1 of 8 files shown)


LibTrixi.jl/lib/build.jl
```diff
@@ -43,7 +43,7 @@ lib_name = "trixi"
 incremental = false
 
 # Do not include stdlibs which are not needed
-filter_stdlibs = true
+filter_stdlibs = false
```
sloede, line 46: Wouldn't the "correct" fix be to add `Random` explicitly to the `Project.toml`?

sloede (approved):
LGTM!

PR #276, opened 2026-02-11
OrdinaryDiffEq updates

benegee:
This PR originates from: https://github.com/trixi-framework/libtrixi/actions/runs/21808243190/job/62915273766#step:19:211

According to the docs (https://docs.sciml.ai/SciMLBase/stable/interfaces/Solutions/#SciMLBase.ReturnCode) the return code has be checked for success differently, and this is what I did.

I am not sure which version actually introduced the change which made the old check fail, nor do I know which version introduced support for the new check. 
Instead, I decided to move to the split OrdinaryDiffEq packages, which I wanted to do at some point anyway.

Doing so, I realized we actually do not need OrdinaryDiffEq in LibTrixi.jl itself (just like with Trixi.jl).

Remaining questions @sloede 
1. Was there a reason to include OrdinaryDiffEq in LibTrixi.jl, which I do not remember?
2. Are the precompilation setting https://github.com/trixi-framework/libtrixi/blob/f5fffee06990fe008cf00fd2fc5214798974c478/LibTrixi.jl/test/Project.toml#L10 obsolete when using OrdinaryDiffEqLowStorageRK only?

### DIFF
--- a/.github/workflows/SpellCheck.yml
+++ b/.github/workflows/SpellCheck.yml
@@ -10,4 +10,4 @@ jobs:
       - name: Checkout Actions Repository
         uses: actions/checkout@v6
       - name: Check spelling
-        uses: crate-ci/typos@v1.42.1
+        uses: crate-ci/typos@v1.43.3
--- a/LibTrixi.jl/Project.toml
+++ b/LibTrixi.jl/Project.toml
@@ -13,7 +13,7 @@ Trixi = "a7f1ee26-1774-49b1-8366-f1abc58fbfcb"
 MPI = "0.20.13"
 OrdinaryDiffEq = "6.53.2"
 Pkg = "1.8"
-Trixi = "0.11.12, 0.12, 0.13, 0.14, 0.15"
+Trixi = "0.11.12, 0.12, 0.13, 0.14"
 julia = "1.8"
 
 [preferences.OrdinaryDiffEq]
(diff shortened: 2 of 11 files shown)


sloede:
> * Was there a reason to include OrdinaryDiffEq in LibTrixi.jl, which I do not remember?

Not 100% sure to be honest 😅. I think it was such that you have all packages you might need installed - how would you do time integration from C/Fortran instead?

> * Are the precompilation setting https://github.com/trixi-framework/libtrixi/blob/f5fffee06990fe008cf00fd2fc5214798974c478/LibTrixi.jl/test/Project.toml#L10
>    obsolete when using OrdinaryDiffEqLowStorageRK only?

Yes, but I am not sure if there are similar preferences for the split packages (maybe ask in the Trixi channel or at the next meeting)

benegee:
> Not 100% sure to be honest 😅. I think it was such that you have all packages you might need installed - how would you do time integration from C/Fortran instead?

Our `libtrixi-init-julia` script will always create a Julia project and add OrdinaryDiffEq* there, just like for the recommended `run` directory in Trixi. 
But I am absolutely fine to have this as a direct dependency in the package itself. 

> Yes, but I am not sure if there are similar preferences for the split packages (maybe ask in the Trixi channel or at the next meeting)

As discussed on slack, there is no similar setting for split packages.

sloede (approved):
LGTM!

PR #299, opened 2026-06-22
Introduced CMake options for the Fortran-API and compiler warnings

Niklas997:
Thank you for developing libtrixi! I wanted to use it with t8code and only needed the C-interface, therefore, I added some CMake-Options to exclude building the Fortran-interface. Additionally, the `-Wall -Wextra -Werror` options have been set obligatory, which has led to errors for me (Ubuntu 24.04. (WSL2), GCC-Version: 13.3.0) due to some unused parameters in the garbage-collection-interface, e.g. "~/.julia/juliaup/julia-1.12.2+0.x64.linux.gnu/include/julia/gc-interface.h:243:47: warning: unused parameter ‘parent’ [-Wunused-parameter]", which is why I added some switches for those options as well. 

The following options with the given default values have been introduced: `LIBTRIXI_BUILD_FORTRAN_API=ON`; `LIBTRIXI_WITH_T8CODE=OFF`; `LIBTRIXI_ENABLE_WARNINGS=OFF`; `LIBTRIXI_ENABLE_WERROR=OFF`; `LIBTRIXI_ENABLE_TESTING=OFF`
Maybe the changes are interesting for libtrixi, but I am not sure, if I covered everything and if that is the intended way for libtrixi.


examples/CMakeLists.txt
```diff
@@ -45,16 +50,18 @@ foreach ( EXAMPLE ${EXAMPLES} )
         target_link_libraries( ${TARGET_NAME} PRIVATE ${PROJECT_NAME}_tls )
     endif()
     if ( T8CODE_FOUND )
-        target_link_libraries( ${TARGET_NAME} PRIVATE T8CODE::T8 )
+        target_link_libraries( ${TARGET_NAME} PUBLIC T8CODE::T8 )
```
benegee: Why is this necessary?

CMakeLists.txt
```diff
@@ -123,12 +149,18 @@ else()
...
         src/auxiliary.h
         src/auxiliary.c
         src/trixi.h
     )
 
+    # Build Fortran library target on demand
+    if( LIBTRIXI_BUILD_FORTRAN_API )
+        target_sources (${PROJECT_NAME} PRIVATE
+            src/api.f90
+        )
```
benegee, line 154: Does PRIVATE result in the same situation as before (when adding the file in den `all_library` call)?

README.md
```diff
@@ -56,7 +56,11 @@ For building, `cmake` and its typical workflow is used.
     - Specifying the directory `install_directory` for later installation is optional.
     - Optional specification of build type sets some default compiler options for optimized
       or debug code.
-    - Building with t8code support is optional. It requires to pass
+    - By default the Fortran-Interface is built; it may be excluded with `DLIBTRIXI_BUILD_FORTRAN_API=OFF`.
```
benegee, line 59: Great, this is indeed a useful option!

README.md
```diff
@@ -56,7 +56,11 @@ For building, `cmake` and its typical workflow is used.
...
+    - Optional compiler options `-Wall -Wextra` are specified with `-DLIBTRIXI_ENABLE_WARNINGS=ON`.
+    - Optional compiler option `-Werror` is specified with `-DLIBTRIXI_ENABLE_WERROR=ON`.
+    - Building of the test is enabled with `LIBTRIXI_ENABLE_TESTING=ON`.
```
benegee: Good idea to unify the naming of the options and adding the LIBTRIXI_ prefix.

examples/CMakeLists.txt
```diff
@@ -45,16 +50,18 @@ foreach ( EXAMPLE ${EXAMPLES} )
...
     target_include_directories(
         ${TARGET_NAME}
         PRIVATE ${CMAKE_SOURCE_DIR}/src
     )
-    if ( T8CODE_FOUND )
-        target_include_directories( ${TARGET_NAME} PRIVATE ${T8CODE_ROOT}/include/t8_fortran_interface/ )
+
+    if ( LIBTRIXI_BUILD_FORTRAN_API AND T8CODE_FOUND )
+        # Add the t8code include directory of the Fortran module file for the Fortran compiler to find the module files 
+        target_include_directories( ${TARGET_NAME} PUBLIC ${T8CODE_DIR}/../include/t8_fortran_interface/ )
```
benegee: Where does the .. come from?

README.md
```diff
@@ -56,7 +56,11 @@ For building, `cmake` and its typical workflow is used.
...
+    - Building with t8code support is optional. It requires to pass `-DLIBTRIXI_WITH_T8CODE=ON` and
```
benegee: Is this approach, i.e. an option DLIBTRIXI_WITH_T8CODE and the using T8CODE_DIR more standard, or cmakish, than what we currently do with T8CODE_ROOT?

benegee (changes requested):
Thanks a lot for your contribution @Niklas997 !

I like the added options! About the warning I am not sure yet.

examples/CMakeLists.txt
Niklas997: I think it is not necessary, and PRIVATE should be preferred.

CMakeLists.txt
Niklas997: Yes, this will result in the same situation.

examples/CMakeLists.txt
Niklas997: This was apparently just a mistake, thanks for pointing that out. I will correct it.

README.md
Niklas997: I am not a CMake-expert, therefore, I do not know which would be the "standard" way. What would you prefer? Should we revert it to use only T8CODE_ROOT?

vchuravy: I think T8CODE_ROOT would be more my expectation from CMake

benegee: I'm not an expert, either. I just read https://cmake.org/cmake/help/latest/variable/PackageName_ROOT.html and like it.

README.md
```diff
@@ -56,6 +56,10 @@ For building, `cmake` and its typical workflow is used.
     - Specifying the directory `install_directory` for later installation is optional.
     - Optional specification of build type sets some default compiler options for optimized
       or debug code.
+    - By default the Fortran-Interface is built; it may be excluded with `DLIBTRIXI_BUILD_FORTRAN_API=OFF`.
+    - Optional compiler options `-Wall -Wextra` are specified with `-DLIBTRIXI_ENABLE_WARNINGS=ON`.
+    - Optional compiler option `-Werror` is specified with `-DLIBTRIXI_ENABLE_WERROR=ON`.
```
benegee, line 61: So far, we have been very strict with respect to warning, because we could. I would thus like to keep it this way, or at least have them on by default.

Regarding the Julia 1.12 issue: https://github.com/trixi-framework/libtrixi/pull/259#issuecomment-3384324273
Currently I would set -DCMAKE_C_FLAGS="-Wno-error=unused-parameter" when running cmake.

benegee (commented):
Oh no, I just realized one of my comments was still pending in github. @Niklas997

README.md
benegee: Any strong opinions on this @sloede ?

CMakeLists.txt
```diff
@@ -138,8 +163,6 @@ else()
     # Version info for the shared object
     set_target_properties ( ${PROJECT_NAME} PROPERTIES SOVERSION ${PROJECT_VERSION_MAJOR} )
 
-
-
     # Include directories
     target_include_directories( ${PROJECT_NAME} PRIVATE src ${JULIA_INCLUDE_DIRS} )
```
sloede:
```suggestion
# Include directories
# PRIVATE: Do not include internal directory in projects building against libtrixi
# SYSTEM: Disable warnings/error from upstream libraries
target_include_directories( ${PROJECT_NAME} PRIVATE src )
target_include_directories( ${PROJECT_NAME} SYSTEM PRIVATE ${JULIA_INCLUDE_DIRS} )
```
Instead of disabling warnings/errors, I recommend to just mark the Julia include dirs as "SYSTEM" includes such that GCC/Clang suppress warnings on them (see also the [CMake docs](https://cmake.org/cmake/help/latest/command/target_include_directories.html).

If this works as expected, I'd also recommend reverting the option of disabling warnings/errors: In practice this leads too often to issues that remain unfixed, and if the "SYSTEM" switch suppresses the warning, we can remove a lot of code complexity from the build system again

README.md
sloede: See my reply above

CMakeLists.txt
```diff
@@ -23,13 +24,26 @@ file(WRITE "${CMAKE_BINARY_DIR}/LIBTRIXI_VERSION" "${LIBTRIXI_VERSION}\n")
 project ( trixi VERSION ${LIBTRIXI_VERSION_MAJOR}.${LIBTRIXI_VERSION_MINOR}.${LIBTRIXI_VERSION_PATCH}
           DESCRIPTION "Interface library for using Trixi.jl from C/C++/Fortran" )
 
-# Enable C and Fortran
-enable_language(C Fortran)
+# Additional configuration options
+option( LIBTRIXI_BUILD_FORTRAN_API "Enable the build of the Fortran interface" ON )
```
sloede: If this is added as an option, you could think about adding also a test - or at least a smoke test - to ensure that builds still succeed with the Fortran interface turned off (I don't think running tests is necessary). It's not super high priority though, so you might also skip the tests 🤷‍♂️

sloede (changes requested):

CMakeLists.txt
```diff
@@ -148,42 +171,52 @@ else()
...
     # Require Fortran 2018 standard for Fortran files
     target_compile_options( ${PROJECT_NAME} PRIVATE $<$<COMPILE_LANGUAGE:Fortran>:-std=f2018>)
 
-
-
     # Add auxiliary *object* library to support fast thread-local storage (TLS)
     add_library ( ${PROJECT_NAME}_tls OBJECT
         src/tls.c
     )
     target_include_directories( ${PROJECT_NAME}_tls PRIVATE ${JULIA_INCLUDE_DIRS} )
```
sloede:
```suggestion
    target_include_directories( ${PROJECT_NAME}_tls SYSTEM PRIVATE ${JULIA_INCLUDE_DIRS} ))
```
As above

PR #302, opened 2026-07-03
Check macOS support branch

benegee:


.github/workflows/Documenter.yml
```diff
@@ -71,6 +71,9 @@ jobs:
           cd docs/doxygen
           doxygen
       - name: Deploy doxygen docs
+        if: >-
+          github.event_name != 'pull_request' ||
+          github.event.pull_request.head.repo.full_name == github.repository
```
sloede:
```suggestion
        # Only deploy on merge to `main`
        if: >-
          github.event_name != 'pull_request' ||
          github.event.pull_request.head.repo.full_name == github.repository
```
For clarity?

examples/trixi_controller_t8code.c
```diff
@@ -22,7 +23,10 @@ void t8_print_forest_information (t8_forest_t forest)
     global_num_elements = t8_forest_get_global_num_elements (forest);
 
     printf ("\n*** T8code ***  Local number of elements:\t%i\n", local_num_elements);
-    printf ("*** T8code ***  Global number of elements:\t%li\n", global_num_elements);
+
+    // Using PRId64 to print t8_gloidx_t portably across platforms
```
sloede: This comment would be more helpful if it included what PRId64 is and where it comes from, e.g., something like

```suggestion
    // Using special integer type PRId64 (defined in t8code.h) to print t8_gloidx_t portably across platforms
```

PR #304, opened 2026-08-06
Update Trixi.jl, including t8code v4 July release

benegee:


.github/workflows/ci.yml
```diff
@@ -93,7 +93,7 @@ jobs:
           T8CODE_RELEASE=${{ matrix.t8code_version }}
           mkdir t8code-local
           cd t8code-local
-          wget https://github.com/DLR-AMR/t8code/releases/download/v${T8CODE_RELEASE}/T8CODE-${T8CODE_RELEASE}-Source.tar.gz
+          wget https://github.com/DLR-AMR/t8code/releases/download/v${T8CODE_RELEASE}-26.07/T8CODE-${T8CODE_RELEASE}-Source.tar.gz
```
sloede: This IMHO defeats the purpose of having a t8code version specified at the top. I recommend to wrap this in either a) an automatic variable, or b) specify the version incl date as the variable above

.github/workflows/ci.yml
```diff
@@ -112,6 +114,9 @@ jobs:
           make -j 2
           make install
 
+          # temporary workaround
+          mv ../prefix/PRIVATE_HEADER/* ../prefix/include/
```
sloede: I'd reference an issue (either here or in t8code) such that your future self (or someone else) knows why this is necessary and when it may be OK to remove again

